PR #455, opened 2012-06-01
Fix config.ru

wagenet:


SteveBenner:
For the sake of posterity, please note that it is officially recommended ***NEVER*** to use

```ruby
require 'rubygems’
```

An explanation for this can be found [on the RubyGems website](http://guides.rubygems.org/patterns/#requiring-rubygems).

PR #560, opened 2012-08-08
Make Rack comparable to middleman server

tdreyno:
An attempt to refactor our internals so Rack can also have access to the file watcher and reloading.


amirnissim:
upgrading to Middleman 3.4.0 required updating `config.ru` as instructed [here](http://www.webuildinternet.com/2015/07/15/load-middleman-using-pow-cx/)
```ruby
require 'rubygems'
require 'bundler/setup'
require 'middleman-core/load_paths'

Middleman.setup_load_paths

require 'middleman-core'
require 'middleman-core/preview_server'

class Middleman::PreviewServer
  def self.preview_in_rack
    @server_information = ServerInformation.new
    @options = {  }
    @app = new_app
    start_file_watcher
  end
end

Middleman::PreviewServer.preview_in_rack
run Middleman::PreviewServer.app.class.to_rack_app
```
with 
```ruby
@options = { latency: 0.25  }
```

PR #1293, opened 2014-06-06
Environments

tdreyno:
This PR creates the concept of an Application `mode` which can be either `:server` or `:build`. This frees `environment` to be able to represent a targeted set of functionality.

The default environment is `:development`. This can be overridden on the CLI with:

```
middleman build -e production
middleman server -e production
```

`config.rb` is loaded as usual and applies to all environments. As before, per-environment config can be set in `config.rb` using:

```
configure :production do
  activate :minify_css
end
```

However, now Middleman optionally loads a corresponding config file at: `environments/production.rb`.

Also included in this PR is a slight tweak to auto-activating extensions to let them only activate in specific modes. For example, we don't replace exceptions with an HTML error page in build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 master
 ===
 
+* Add support for `environments` with the `-e` CLI flag. Loads additional config from `environments/envname.rb`. Removed `development?` helper in favor of `environment?(:development)`. Added `server?` helper to differentiate between build and server mode.
 * Removed `with_layout`. Use loops of `page` instead.
 * Removed Queryable Sitemap API
 * Removed `css_compressor` setting, use `activate :minify_css, :compressor =>` instead.
--- a/middleman-cli/lib/middleman-cli/build.rb
+++ b/middleman-cli/lib/middleman-cli/build.rb
@@ -18,6 +18,10 @@ module Middleman::Cli
     namespace :build
 
     desc 'build [options]', 'Builds the static site for deployment'
+    method_option :environment,
+                  aliases: '-e',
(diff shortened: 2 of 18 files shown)


vasco:
I'm not able to get this to work (tried with 3.3.7, v4.0.0.alpha.6 with jruby).

Running: `bundle exec middleman server -e production`

The `#environment` method, correctly returns `:production`, however:
* `environments/production.rb` isn't loaded, and
* the following code raises `NoMethodError: undefined method 'production_config' for Middleman::Application::MiddlemanApplication1:Class`
```ruby
configure :production do
 # ...
end
```

Am I doing something wrong?

sandstrom:
@vasco Version 3.3.7 of Middleman? I don't think it's in the v3 branch, only v4. If you haven't, try with `master` and see if it works.

vasco:
@sandstrom This PR didn't land in middleman's 3.3.7 version. When I first wrote my comment, I was wrongly getting v4, I fixed that and I can now confirm that, with middleman version 4.0.0.alpha.6, everything works.

I'm doing: `gem "middleman", "~> 4.0.0.alpha.6"`.

tdreyno:
Yup, this is in v4. That said... "alpha" release so be careful.

equinusocio:
How i can set multiple environments through rb files? What i have to put inside this?
```
environments/development.rb
environments/build.rb
```

sandstrom:
@equinusocio here are the docs: https://middlemanapp.com/advanced/configuration/#environment-specific-settings

*(this question is better suited for Stackoverflow, btw)*

PR #1321, opened 2014-07-14
Add external command support

tdreyno:
Okay, this is still a work in progress, but I want to chat about it.

This is a thin API for controlling external processes inside Middleman. Basically, it let's you boot Grunt in watch/build mode and use that to feed content into the sitemap. The API looks like (using Ember's Broccoli CLI as an example):

```
activate :external_pipeline,
  name: :ember,
  command: "cd test-app/ && ember #{build? ? :build : :serve} --environment #{config[:environment]}",
  source: "test-app/dist"
```

We start the command, async in server mode, sync in build mode. Then we merge the results of some temp folder into the sitemap (BTW, I'm using this feature to explore a simpler API for "multiple source directories" which basically just means merging multiple `OnDisk` sitemap instances.


shaekuronen:
What's the status on this?  Looks like at least some of it got merged into master, but not seeing any documentation?

Put snippet above in config.rb and get `Unknown extension: external_pipeline` error...

tdreyno:
It's in v4

sandstrom:
Any updates on v4? :smile:

tdreyno:
@sandstrom Got it into production on a major site last week :) Realized we removed a feature that a surprising number of extensions use, so going to add it back before beta.

That said, safe to experiment with GH master branch if you want. Will be very similar to beta 1.

sandstrom:
Sounds great! :) I'll try it out!

PR #1335, opened 2014-07-28
Allow middleman.rb as an alternative to config.rb.

tdreyno:
This is for discussing renaming the base config for v4. Also added the ability for the Git project init to work on specific branches.


marcinant:
Please reopen this issue.
This is not solved and commit merged on 15 Aug 2014 doesn't contain anything related to this issue.

PR #1386, opened 2014-10-03
Parsing HTML with regular expressions

AndrewKvalheim:
I've hit the bug illustrated by this failing test in which `' + generate() + '` is fed into Sass.

I'm not terribly excited about patching [this regex][regex]; what are your thoughts on using Nokogiri/[Oga][replace] here instead? How much of a concern is performance?

  [regex]: https://github.com/AndrewKvalheim/middleman/blob/429e7d64bd306c5139748ad578c47616bc4dfef2/middleman-core/lib/middleman-more/extensions/minify_css.rb#L32
  [replace]: https://github.com/middleman/middleman-blog/pull/226

### DIFF
--- a/middleman-core/features/minify_css.feature
+++ b/middleman-core/features/minify_css.feature
@@ -106,9 +106,7 @@ Feature: Minify CSS
     When I go to "/inline-css.html"
     Then I should see:
     """
-    <style>
-      Hello
-    </style>
+    <style>Hello</style>
     """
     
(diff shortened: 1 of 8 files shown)


AndrewKvalheim:
This is working now that YorickPeterse/oga#56 is resolved.

tdreyno:
I don't think we want to do this for v3-stable, can you re-target at master and I'll merge it in

PR #1415, opened 2014-12-28
Update version of i18n

ghost:
I want to embed my middleman-extension `middleman-presentation` into a rails-app. Rails requires `i18n` to be `0.7.0`. `middleman` requires it to be `~>0.6.9`. This PR upgrades the version. 

I ran the test suite and no test failed because of this upgrade. When running `git diff v0.6.9..v07.0` on the the [repository](https://github.com/svenfuchs/i18n/compare/v0.6.9...v0.7.0) it looks more like a minor release.

 Along with this change I discovered some minor `rubocop`-offenses I will fix in a separate PR. I would be happy, if you merge this change and release a new minor version with it. :-) Thanks.

### DIFF
--- a/middleman-core/middleman-core.gemspec
+++ b/middleman-core/middleman-core.gemspec
@@ -39,5 +39,5 @@ Gem::Specification.new do |s|
   s.add_dependency("listen", [">= 2.7.9", "< 3.0"])
 
   # i18n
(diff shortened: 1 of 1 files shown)


dg-ratiodata:
It's quite a mess. :-) There are several gems which needs an update. Working on that.

PR #1416, opened 2014-12-28
Fix rubocop offenses

dg-ratiodata:
I "discovered" some rubocop offenses. This PR fixes them.


dg-ratiodata:
> Does this code need renaming AccessorMethodName → Style/AccessorMethodName, with a second comment below it like yours?

Yes, of course. Fixed that. I also fixed some other offenses.

PR #1427, opened 2015-01-15
add specs for #1426 & reorganize Middleman::Util specs to a single file

stevenosloan:
Adding the specs I should have included with #1426, ~~and then tracked down the errors caused by [this](https://github.com/cucumber/aruba/commit/5b2c7b445bc80083e577b793e4411887ef295660) update to aruba.~~ (jk, you already did that)

Also "cleaned up" the specs dir a bit -- so all util specs are at `/spec/middleman-core/util_spec` instead of split across a few files.


tdreyno:
Nice, thanks @stevenosloan

PR #1444, opened 2015-02-10
Add :overwrite option to gzip

joost:
Another suggestion :)
Useful for deployment to S3 together with gzip content-encoding.

### DIFF
--- a/middleman-core/lib/middleman-core/extensions/gzip.rb
+++ b/middleman-core/lib/middleman-core/extensions/gzip.rb
@@ -12,6 +12,7 @@
 class Middleman::Extensions::Gzip < ::Middleman::Extension
   option :exts, %w(.js .css .html .htm), 'File extensions to Gzip when building.'
   option :ignore, [], 'Patterns to avoid gzipping'
+  option :overwrite, false, 'Overwrite original files instead of adding .gz extension.'
 
   class NumberHelpers
     include ::Padrino::Helpers::NumberHelpers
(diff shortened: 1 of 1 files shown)


tdreyno:
Sounds like a good idea. Thanks!

monfresh:
I was looking for details about the gzip extension in the documentation and couldn't find much. I was particularly interested in the overwrite option and the only way I found out about it was doing a search for "gzip" in the issues and came across this PR.

It would be great to add documentation for this feature as it's not clear to me how to specify it. I tried the following, but neither worked:
```
configure :build do
  activate :gzip, :overwrite
end
```

```
configure :build do
  activate :gzip do |gzip|
    gzip.overwrite
  end
end
```
How can I specify the overwrite option? Thanks!

Arcovion:
`activate :gzip, overwrite: true`

monfresh:
I tried that too, and it did not work. I got the following error:
```
/Users/moncef/.rvm/gems/ruby-2.2.3@middleman/gems/middleman-core-3.4.0/lib/middleman-core/configuration.rb:157:in `define_setting': Setting overwrite doesn't exist (RuntimeError)
```
Is this only in 4.0?

Arcovion:
Ah, yea it's only in v4, was not backported to v3. Explains why it's not in the docs yet too.
I would personally just monkeypatch this in the meantime... like this would overwrite stuff:
```ruby
class Middleman::Extensions::Gzip
  def gzip_file(path)
    input_file = File.open(path, 'rb').read
    output_filename = path + '.gz'
    input_file_time = File.mtime(path)
    File.open(output_filename, 'wb') do |f|
      gz = Zlib::GzipWriter.new(f, Zlib::BEST_COMPRESSION)
      gz.mtime = input_file_time.to_i
      gz.write input_file
      gz.close
    end
    File.utime(File.atime(output_filename), input_file_time, output_filename)
    old_size = File.size(path)
    new_size = File.size(output_filename)
    [output_filename, old_size, new_size]
  end
end
```

PR #1445, opened 2015-02-10
Add Woff2 to asset hash extensions

alex88:
I had some 404 on my site and it was an issue with font-awesome woff2 file that didn't get the hash

### DIFF
--- a/middleman-core/lib/middleman-core/builder.rb
+++ b/middleman-core/lib/middleman-core/builder.rb
@@ -16,7 +16,7 @@ module Middleman
     def_delegator :@app, :logger
 
     # Sort order, images, fonts, js/css and finally everything else.
-    SORT_ORDER = %w(.png .jpeg .jpg .gif .bmp .svg .svgz .webp .ico .woff .otf .ttf .eot .js .css)
+    SORT_ORDER = %w(.png .jpeg .jpg .gif .bmp .svg .svgz .webp .ico .woff .woff2 .otf .ttf .eot .js .css)
 
     # Create a new Builder instance.
     # @param [Middleman::Application] app The app to build.
--- a/middleman-core/lib/middleman-core/extensions/asset_hash.rb
+++ b/middleman-core/lib/middleman-core/extensions/asset_hash.rb
@@ -1,7 +1,7 @@
 require 'middleman-core/util'
 
 class Middleman::Extensions::AssetHash < ::Middleman::Extension
-  option :exts, %w(.jpg .jpeg .png .gif .webp .js .css .otf .woff .eot .ttf .svg), 'List of extensions that get asset hashes appended to them.'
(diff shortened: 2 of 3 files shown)


alex88:
Let me know if it could be merged so I'll squash it

tdreyno:
Thanks. Are there any other places we mention `.woff` in the codebase?

alex88:
Ok this should be the last one!

alex88:
There is only a strange thing happening on my local setup and you can see it here http://alexfu.it.s3-website-eu-west-1.amazonaws.com/

The woff2 file saved in the fonts folder has an hash digest of `10752316` however it gets linked as `56ce13e7` which is the same as the woff version, coincidence? I've tried also with an empty build folder, same thing.

tdreyno:
Seems suspicious. Are the source files the same content?

Maybe it's coming from Sprockets: https://github.com/middleman/middleman-sprockets/blob/95e9841a7e036ab9a25f86b13909b2d593128eb3/lib/middleman-sprockets/asset.rb#L110-L112

alex88:
Oh could be, let me try to monkey patch that too and see if it needs a PR

alex88:
Btw, no it's not the same content, as the files are created with a different hash, it's just the linking in the stylesheet that's wrong

alex88:
Anyway, seems that even monkey patching `is_font_by_extension?` the file gets linked with the same asset hash, maybe there's some kind of regex somewhere that matches both woff and woff2 and catches the first one?

alex88:
Ok, I've found the "workaround":
Basically after the monkey patches (middleman core and sprockets):
- including font-awesome-sass gem directly via sprockets `*= require font-awesome`
  - in development it was that linking `fonts/font-awesome/fontawesome-webfont-56ce13e7.woff2` but the actual file it was serving `fonts/font-awesome/fontawesome-webfont-10752316.woff2`
  - in build it was building the correct `10752316` file but linking the `56ce13e7` (which refers to the woff version)
- including it via sass imports and so using `font-path` due https://github.com/FortAwesome/font-awesome-sass/blob/master/assets/stylesheets/_font-awesome-sprockets.scss and https://github.com/FortAwesome/font-awesome-sass/blob/master/assets/stylesheets/font-awesome/_path.scss made it working fine, the generated and linked hash were the same

alex88:
Any chance to get this merged?

maxmeyer:
@tdreyno 

I think  https://github.com/middleman/middleman-sprockets/blob/95e9841a7e036ab9a25f86b13909b2d593128eb3/lib/middleman-sprockets/asset.rb#L110-L112 needs to be changed too.

alex88:
@maxmeyer the fix is already there in master https://github.com/middleman/middleman-sprockets/blob/master/lib/middleman-sprockets/asset.rb#L111 I've done the change in https://github.com/middleman/middleman-sprockets/pull/76

maxmeyer:
ok. didn't see that. thx.

tdreyno:
Just pushed out the new middleman-sprockets gem

PR #1457, opened 2015-02-18
Fix slim >= 3.0.0 deprecation warning

Arcovion:


tdreyno:
What happens on slim < 3?

Arcovion:
It breaks, though I think all Slim users should be on 3.0 by now; no breaking changes were done recently apart from dropping Ruby 1.8.7 as far as I can tell.
Didn't know it was already patched in v4.

Issue #1390, opened 2014-10-08
I18n: Localized paths helper and frontmatter keys


joallard:
I've done a bit of reading in the issues, and I think it's time I bring up my pet enhancement in Middleman, which is out-of-the-box link helpers for localized paths. I even want to go a step further, which is to make it work without putting the paths in the dictionary.

We've discussed refactoring in depth in #924, but I think that this enhancement warrants its own issue as I18n is a bit of a mega-feature. Multiple people also seem to have raised such a need, such as #1146, #1306.

Here are the end goals I see:
- Have the possibility to easily link to localized content in a language-agnostic way, ie. in a layout nav
- Install a locale switcher link, which would link to the same resource, if available, in the target locale
- Not relying on the locale dictionary to make this work

The alternative I'd suggest to putting paths in the locale file is this: Put a frontmatter key on localized pages that would identify pages. If no key is there, the extensionless path of the page is assumed.

In Gherkin terms, here's what I have in mind:

``` gherkin
    Given an i18n fixture app
    And a file named "config.rb" with:
      """
        activate :18n
      """
    And a locale dictionary like:
      """
        en:
          about: About
        fr:
          about: À propos
      """
    And a layout file with:
      """
        <%= link_to t(:about), locale_path(:about) %>
      """
    And a file named "about.en.html.erb" with:
      """
        Here's what it's all about
      """
    And a file named "a-propos.fr.html.erb" with:
      """
        ---
        key: about
        ---
        Voici ce que nous faisons
      """
    Given the server is running
    When I go to "/"
    And I click on "About"
    Then I should see "Here's what it's all about"
    When I go to "/fr/"
    And I click on "À propos"
    Then I should see "Voici ce que nous faisons"
```

I realize `key` is probably not the ideal... uh... key.

In code terms, the `localization_data` hash seems like the perfect place to store this kind of info.

I'd contribute it myself, I just need to know if my approach is the one we want to take. I'm also a bit disoriented about where to put my fixtures and tests, but nothing impossible!

Time for feedback!

sandstrom:
I'm doing some issue-gardening 🌿🌷🌾 and came across this issue.

Sorry that this has lingered for quite a while! I'd suggest either closing this issue or submitting a PR with an implementation.

Dalpatgohil115:
Set it

Issue #924, opened 2013-06-04
Refactor i18n


tdreyno:
This is a proposal for refactoring how i18n works in Middleman v4.

First, we remove the magical `localizable` folder.

When `i18n` is activated, all html files in source are localized by default into as many languages as the user specifies. These files are output into their own folders (en, es, fr) as usual.

To be more specific, there are additional options:

Naming a file with the locale will target only that specific output folder. For example: `about.fr.html.erb` will only output to `fr`.

Adding `localizable: false` to the front matter will not move the file during build and will have the default locale set.

Adding a glob or a proc to the activation can mass skip localization:

```
activate :i18n do |i|
  i.skip = "/legal/*"
end
```

Questions:

* Do non-localized files output to the root or to every single locale folder, but they are the same?
* Does any of this seem natural to Rails users?
* Do non-html files need to be localized? Strings in JS?
* Does this help or hinder localizing blogs?

Tracking:

- [ ] Remove `localizable` folder
- [ ] Add `localized` front matter
- [ ] Add `filter` to extension to specify which files in `source` are localized.
- [ ] Make URL helpers (`link_to`, etc) aware of `i18n`

PR #1466, opened 2015-02-21
I18n: Switch locale helpers

joallard:
This is as a follow-up to #1390 and #924.

> - Have the possibility to easily link to localized content in a language-agnostic way, ie. in a layout nav
> - Install a locale switcher link, which would link to the same resource, if available, in the target locale
> - Not relying on the locale dictionary to make this work

In practice, the following helpers are being added:
- [x] ~~`locale_path(page_id, locale = I18n.locale)`, so we can easily construct eg. a `<nav>` with language-agnostic keys.~~ (integrated into `url_for(*args, locale:)` and thus `link_to`)
- [x] `other_locales` and `other_locale`, some niceties for the following method...
- [ ] `switch_locale_path(locale = other_locale)`, so we can easily switch language while staying on the same resource

Internally,

I recycled `@maps` that seemed to be useless and used it as a localization map, it would now look like this:

``` rb
{"president"=>{:en=>"images/president.svg", :es=>"es/images/president.svg"},
 "hello"=>{:en=>"hello.html", :es=>"es/hola.html"},
 "flag"=>{:en=>"images/flag.svg", :es=>"es/images/flag.svg"},
 "index"=>{:en=>"index.html", :es=>"es/index.html"},
 "morning"=>{:es=>"es/manana.html", :en=>"morning.html"},
 "one"=>{:en=>"one.html", :es=>"es/una.html"},
 "partials.index"=>{:en=>"partials/index.html", :es=>"es/partials/index.html"}}
```

Which makes it easy to find a localized path.

The extension is in bad need of a refactor. I can do some things if the feedback is positive.

I've had to prioritize loading the Frontmatter extension before I18n, because we're using the `page_id` frontmatter key. Since Routing depends on I18n, this means Frontmatter is now before Routing as opposed to after. Don't know what this means for the rest of the features.

```
Failing Scenarios:
cucumber features/frontmatter_page_settings.feature:35 # Scenario: Overriding layout through frontmatter
```

Work in progress; please comment.


tdreyno:
Thanks for diving in @joallard. I'm traveling right now, but will try to review and comment soon.

tdreyno:
You're absolutely right, this extension is really crufty. I've merged in a lot of features, not knowing which are good or bad because I only use i18n in the simplest way, myself.

We'd love any direction, code and ideas you have.

I really like this iterative approach, slowly adding features rather than a complete do-over.

I'm going to push out v4 beta 1 today, so let's save this for beta 2. And we need to track down that failing test.

joallard:
Awesome. I'll go forward then!

joallard:
I can't find where the layout option has been implemented. The tests have been added after the implementation. Pointers?

tdreyno:
Should be this: https://github.com/middleman/middleman/blob/master/middleman-core/lib/middleman-core/core_extensions/front_matter.rb#L44-L45

I think it's unrelated to i18n, but it is related to changing the order things run in.

joallard:
About that failing test: it's because Routing overrides the Frontmatter layout setting with what's in `config.rb`. I changed the order to `{Routing: 60, FrontMatter: 70, I18n: 80}`, but `features/capture_html.feature` started to fail. I don't know what to do.

joallard:
Can I get help on this?

tdreyno:
Haven't had time. `capture_html` and all the Padrino internals are super gnarly and very prone to breakage. Lets change it so `capture_html` fails and we'll fix it post-merge.

tdreyno:
@joallard Can this be brought up to date with beta2?

joallard:
Okay, let me get to it

tdreyno:
@joallard While you're at it, can you target the `v3-stable` branch?

joallard:
Got it!

joallard:
Can't do `v3-stable`. The diff is too large (it's a v4 extension and I was based on master—plus some differences in how the resources are fetched I think) and I don't understand what's there. I lost the context I had in Feb.

I rebased on master, but there are still some failures.

Lightheaded:
Any updates on this issue?

Lightheaded:
Any updates for the issue?

joallard:
@tdreyno Is it okay if this is a master-only patch? (ie v4 I presume?)

tdreyno:
Yes

middleman-core/lib/middleman-core/core_extensions/i18n.rb
```diff
@@ -89,6 +92,31 @@ def locate_partial(partial_name, try_static=false)
...
+
+    # Path to the same page in another locale
+    def switch_locale_path(locale=other_locale)
+      raise ArgumentError, 'No other locale to switch to' unless locale
+      locale_path(current_page.locals[:page_id], locale)
+    end
+
+    def other_locale
+      other_locales[0]
+    end
```
johnnyshields: This method seems superfluous, is it really needed?

johnnyshields:
Does this PR add support for I18n'ed frontmatter keys (as per the title of the PR)? I don't see it in the code...

joallard:
@tdreyno I'm a little busy at the moment, and my confusion with the failing tests and ext load order (among larger components) has prevented me from being more proactive in fixing this PR, but let me know if there's anything I can do or answer. I'd be happy to help Middleman get a killer i18n any way I can, it's a truly great project you've made you guys.

tdreyno:
Cool. I'll try rebasing and see where things stand now. Thanks!

pixelchef:
Would really love to use this as a clean way to link to locale blog.sources in config.rb to keep articles from different locales separate!

joallard:
9de1f16f

joallard:
Seeing `url_for` has been implemented in parallel, here are the parts that this PR is aiming to:

1. I can link to the same page in another locale (⇒ 2)
2. I can link to another page in an arbitrary locale (⇒ 3)
3. Every page has a `page_id` that is unique, rememberable and independent of locale
4. I can change the `page_id` in a page frontmatter

joallard:
What I'm tiptoeing around is the `page_id` logic. That mechanism isn't really solid, and it seems to me the urls in general should be able to refer to a certain `page_id`, a little à la how Rails has `as` labels for routing. 

(Btw, the rebase fixed the extension priorities that seem to have magically been sorted out)

joallard:
@tdreyno, what's your take on this?

middleman-core/lib/middleman-core/core_extensions/i18n.rb
```diff
@@ -256,10 +273,20 @@ def parse_locale_extension(path)
     [locale, path, basename]
   end
 
-  LocalizedPageDescriptor = Struct.new(:path, :source_path, :locale) do
+  # Turn path into a page_id for the lookup map
+  # eg. "foo/bar/baz.foo.html" => "foo.bar.baz.foo"
```
tdreyno: Why is the actual path string not enough? What does changing this into all dots get us? Is it intended to be an actual structure from a locale file?

tdreyno:
I link an arbitrary `page_id` concept or linking in MM is a really good idea. Would the `page_id` be the same for all translations of the same resource?

That said, this is looking pretty good. Those helpers will be nice.

middleman-core/lib/middleman-core/core_extensions/i18n.rb
joallard: The idea is that it is then easy to use it as a key for lookup into I18n. So one could do: `{foo: {bar: {baz: {foo: "Text"}}}}` However thinking about it we could probably keep the slashes for `page_id` and do the conversion into I18n lookup key later in the process. But that's the idea behind.

joallard:
> Would the page_id be the same for all translations of the same resource?

Exactly. Why it is there is that for localized versions of the same file, eg. `about.en.html` and `a-propos.fr.html`, the filename needs to be different to reflect the url, even if it represents the same resource (in a different language format). The default is to take the filename/path, but we need to adjust that in the other languages so they can be "linked" together. Hence one could put `page_id: about` in the frontmatter of `a-propos.fr.html` and be all set.

I just thought it's a little clunky to tack that logic just in I18n while it could be baked in deeper into Middleman as it answers to how are pages identified, regardless of their language. We could then say let's `link_to t(:about), :about` instead of `link_to t(:about), "/about.html"`.

tdreyno:
@joallard I agree, let's put that into core. It can be a piece of general metadata on a resource, set by `page` or `proxy` and overridden by `frontmatter`. Let me take a crack at that today.

tdreyno:
Okay @joallard, how's this for a nice base: https://github.com/middleman/middleman/commit/edb1a53060abdc3589cbcfb6a60b69796462c999

joallard:
Yep, I'm on it

johnnyshields:
@joallard @tdreyno what's the latest on this one?

joallard:
Rebasing was a little less straightforward than I thought and I ran out of time, but it's on my to-do list, between real life things and maintaining Selectize. I can share my WIP if that helps.

tdreyno:
@joallard WIP is always going, I'm going on vacation next week, I may actually have to to investigate this

joallard:
@tdreyno Not sure I understood this last one

johnnyshields:
@joallard maybe it might be easier to break this into smaller tickets that can be merged one-by-one?

joallard:
Whew! And here we are. Works. Rebased. As we can see (I activated verbose commit mode), it pretty much comes in one piece. It's dependent on #1996, so we'll solve that first.

What's left:
- [ ] That `switch_locale_path` helper
- [ ] Clean up `url_for`
- [ ] Clean up Git history

We'll probably get some edge cases too to add in there.

joallard:
Also, is there a reason we don't use Pry out of the box in dev?

joallard:
Next up: remove the (*) `localizable` folder.

tdreyno:
@joallard I think because we used to support versions of Ruby which didn't support pry... but now, good to include by default

manuelmeurer:
Hi @joallard and @tdreyno, what it the status of this PR? I just ran into the issue that `link_to`/`url_for` with I18n doesn't work as expected (at least with `directory_indexes`, I didn't check without). Will the changes in this PR fix this? What's needed to move forward?

johnnyshields:
FYI here's my I18n helper methods I've added to my Middleman setup: https://gist.github.com/johnnyshields/98a695df51b1e99f3593579d3c9a3fd1

One thing to note is that my primary locale is Japanese, so my site root (/) is Japanese, and other locales are at /en/... etc. Would be good if this can be support in this PR.

joallard:
I had a refactor lingering, it's pushed at 6e89a36. @tdreyno you'll probably notice that I've got references to the `i18n` in my classes everywhere, should this have been structs?

jnf-si:
Hi! I just wanted to take a moment to thank yinz for the ongoing work on this issue. This (year-long!) thread has been really informative and has helped me figure out how to move forward with my Middleman project. I also really appreciate the supportive tone and graceful acknowledgement that OSS is hard, language is hard, and people are busy.

Thanks again for sticking with this. 👍

tdreyno:
Looking for a passing build, but happy to review whenever

joallard:
I don't have so much time to dedicate to this right now, but it should be only a matter of fixing a few loose ends on the tip of the branch. I'm happy to help if someone wants to pick it back up.

tdreyno:
That liquid error is a whole other thing I need to clean up (likely by removing tests for old versions of liquid), you can ignore it

kstratis:
Hi, any update on this one?

dorellang:
I've been hacking Middleman a lot (by monkey patching and reopening some framework classes) to get Internationalization, Blogging and Directory Indexes playing along for my personal website. I think I grasp how the code currently on master works and I'd love to contribute, but I don't know what the project standards are, how the code should be tested, if there's any backwards compatible behavior that should not be broken, etc.

I'd really appreciate any information. I'm looking forward to solving this issue, since the solution I've made for my website just feels like a complete duct tape mess and I'm pretty sure that would break in the next Middleman release. It just would be freaking nice to get it working in a cleaner way.

Cheers!

ChristianPeters:
I also spent a lot of time, having to understand nearly all messy internals to be able to duct tape our locale switch helpers together.
Sadly, I can't find any time to contribute but I could share my helper structure if that might be helpful.

dagumak:
Any updates?

tdreyno:
@dorellang If you can write a specification of how it should work, we can convert that to a Cucumber test

joallard:
This was a good run, y'all, it brings back memories. 😉 Glad you were able to clean this one up @tdreyno

PR #1497, opened 2015-04-16
Revert "Try to avoid *nix double load issue again"

mtsmfm:
rackup doesn't work because config.ru template uses
`Middleman.server` without call `Middleman::Application`

This reverts commit 14104aad70e9431d4891e835474efa7937b396a8.

### DIFF
--- a/middleman-core/lib/middleman-core.rb
+++ b/middleman-core/lib/middleman-core.rb
@@ -8,10 +8,9 @@ $LOAD_PATH.unshift(libdir) unless $LOAD_PATH.include?(libdir)
 module Middleman
   # Backwards compatibility namespace
   module Features; end
-
-  autoload :Application, 'middleman-core/application'
 end
 
 require 'middleman-core/version'
 require 'middleman-core/util'
 require 'middleman-core/extensions'
+require 'middleman-core/application' unless defined?(::Middleman::Application)


tdreyno:
This should be fixed by adding a require to the bits that make `.server` work, not reverting this fix.

tdreyno:
@mtsmfm Try running MM from Git and seeing if things work better.

```
gem "middleman", github: "middleman/middleman", branch: "v3-stable"
```

mtsmfm:
It seems that 60b3bfc doesn't fix this problem. (I tried https://github.com/mtsmfm/middleman_rackup)
`middleman-core/core_extensions/request.rb` is not required if we require only `middleman/rack`.

`middleman-core/core_extensions/request` is required by `middleman-core/core_extensions` only,
and
`middleman-core/core_extensions` is required by `middleman-core/application` only.

tdreyno:
Well, that commit includes a require to application, so it should trickle down.

mtsmfm:
No, 60b3bfc can't fix this issue.

```
$ git clone https://github.com/mtsmfm/middleman_rackup
$ cd middleman_rackup
$ BUNDLE_GEMFILE=Gemfile-mmv3stable bundle install
$ BUNDLE_GEMFILE=Gemfile-mmv3stable bundle exec rackup
/home/mtsmfm/.ghq/github.com/mtsmfm/middleman_rackup/config.ru:4:in `block in <main>': undefined method `server' for Middleman:Module (NoMethodError)
```

config.ru template only [requires `middleman/rack`](https://github.com/middleman/middleman/blob/v3-stable/middleman-core/lib/middleman-core/templates/shared/config.ru#L2).

60b3bfcee2 changes `middleman-core/lib/middleman-core/core_extensions/request.rb`,
but `middleman-core/lib/middleman-core/core_extensions/request.rb` isn't required
when we call `require 'middleman/rack'`.

PR #1498, opened 2015-04-17
allow setting a destination_dir for SourceWatcher(s)

stevenosloan:
This change lets a source be defined with a destination directory. That way we can mount a source in a namespace and worry less about name collisions with files in other sources.

Currently if you mount a source with an `index.html`, the index at `source/index.html` gets overlooked even if they should have different destinations. The use case behind this (that I'm working on at least), would be for gem documentation -- were we're mounting the `gem_name/docs` directory with a SourceWatcher (so those files can be rendered).


tdreyno:
Nice idea.

Issue #1396, opened 2014-10-20
Supporting srcset attribute analogously to src attr in image_tag


komor72:
There is a relatively new [srcset](http://caniuse.com/#feat=srcset) attribute in `img` tag that can be useful for HiDPI/Retina devices and is the second easiest way to support them. To properly use it in Middleman it should be handled analogously to `src` attribute (supporting relative/absolute paths computing). Unfortunately it also means hacking it in Markdown engines and other high-level HTML-abstractions plugged into Middleman.

I'm considering an example with extended syntax using [attribute list definition](http://kramdown.gettalong.org/syntax.html#attribute-list-definitions) from Kramdown:

``` markdown
![Alternative text](/folder-relative-or-absolute/image.jpg "Title attribute"){: srcset="/folder-relative-or-absolute/image@2x.jpg 2x"}
```

This should generate (in case of relative paths enabled and Markdown source 1 folder deeper than image folder):

``` html
<img alt="Alternative text" title="Title attribute" srcset="../folder-relative-or-absolute/image@2x.jpg 2x" src="../folder-relative-or-absolute/image.jpg">
```

The generated path in `srcset` attribute of `img` tag should be computed relative/absolute-wise the same way as `src` is. Currently it’s not, so the `../` part is missing, I only get `/`. To make things more difficult the syntax of `srcset` is quite flexible and contains more than a path, in the simpliest case.

Any hope?

sandstrom:
I'm doing some issue-gardening 🌿🌷🌾 and came across this issue.

Did the suggestion by @tdreyno work? If so, can this be closed?

komor72:
Wow, I haven’t seen **dperrymorrow**’s commit. Will check if it works (in MM 3.4.1) any time soon. Unfortunately I’m a occasional user (and don’t use MM4 in any project yet, at least not untill some crucial gems are compatible).

PR #1500, opened 2015-04-22
Pull request for issue #1396 supporting srcset property in image tags

dperrymorrow:
this adds an extension for ```image_tag``` which enables used of ```srcset``` property in image tags, and having the urls for the paths built using asset_url just like it does for your main src. Leaves absolute
urls alone, _(having // in the path)_

**Erb:**
``` erb
<%= image_tage 'pic_1980.jpg', srcset: 'pic_640.jpg 2x, pic_1024.jpg 3x' %>
```
``` html
<!-- output -->
<img src="/images/pic_1980.jpg" srcset="/images/pic_640.jpg 2x, /images/pic_1024.jpg 3x">
```

Also work for markdown rendering, as it is processed through the ```image_tag``` helper.

**Markdown:***
``` markdown
![image](blank.gif){:srcset="image_2x.jpg 2x"}
```
``` html
<!-- output -->
<img src="/images/blank.gif" srcset="/images/image_2x.jpg 2x">
```


tdreyno:
This is great. I actually don't see any reason not to put this directly in `image_tag` and skip the extension portion. What do you think?

dperrymorrow:
Ok sure, so I should just extend image_tag in the default helpers instead of doing the whole 
``` ruby
class Middleman::Extensions::ImageSrcsetPaths < ::Middleman::Extension
```
bit? If so I fix that up.

tdreyno:
Exactly. Thanks

dperrymorrow:
ok, did that and squashed.

tdreyno:
Lovely, thanks!

PR #1515, opened 2015-05-06
fixed 'after_render' hook is not work.

yterajima:
- 'before_render' hook is called twice.
- add simple cucumber test about some hooks.


tdreyno:
Thanks!

PR #1517, opened 2015-05-07
Added support for complete path localization

splebel:
The full path is now localized by build_resource, this is useful with subfolders, eg: /my/great/page -> /mon/super/site


tdreyno:
Would this have any regressions against users expecting the old functionality?

tdreyno:
Ping @splebel

splebel:
There shouldn't be any regression as I don't see the use case of a half translated url, eg: /my/great/page -> /mon/great/page

PR #1518, opened 2015-05-07
fixed: Builded text file(html, css, xml, txt...)'s permission is 0600

yterajima:
In v3-stable branch, builded text file's permission are 0644.
screenshot 2015-05-07 19 36 08

But in master(v4) branch, file's permission are 0600.
screenshot 2015-05-07 19 40 40

When I deploy(drug-and-drop with sftp), the website isn't displayed. So
I fixed this problem.


tdreyno:
Interesting. Apparently a side effect of some Tempfile stuff, I think. Thanks!

PR #1526, opened 2015-05-19
Allow minify to correctly throw warnings

zamith:
The `path` variable was being used but never defined in the context of the `minify` method. Thus, instead of a warning we would get an error. Using an instance variable fixes this.


tdreyno:
:+1:

PR #1529, opened 2015-05-24
Add rel=canonical URL to default RedirectResource template

rmm5t:
See https://support.google.com/webmasters/answer/139066


tdreyno:
:+1: Great idea.

rmm5t:
Thanks. Do you need a PR against master too, or is there already a process in place for re-syncing with the v3-stable branch?

tdreyno:
It's resync, eventually. I run a merge every so often.

PR #1533, opened 2015-05-28
fix #1496 : Asset hashes on WOFF vs WOFF2 fonts get confused

### LINKED ISSUE #1496: Asset hashes on WOFF vs WOFF2 fonts get confused
**(v3-stable)**

When having both, a font.woff and a font.woff2 file in place/in use, Middleman confuses their hashes.

That's to say, the woff2 font will be referenced  by the hash actually belonging to the woff font.

**Example**

*Actual* file names:
raleway-medium-webfont-46ed8c33.woff
raleway-medium-webfont-7b27610c.woff2

*Referenced* as:
raleway-medium-webfont-46ed8c33.woff
raleway-medium-webfont-46ed8c33.woff2 (doesn't exist)

akarzim:
reverse sorting the extensions regex solves this issue


tdreyno:
Nice detective work!

PR #1540, opened 2015-06-14
Allow extended globs with ignore

Arcovion:
This makes it match the default behaviour of Dir#glob

Edit: Ok, `uninitialized constant File::FNM_EXTGLOB (NameError)` on 1.9.3... will add an if statement.
Edit2: Amended


tdreyno:
:+1:

PR #1555, opened 2015-07-04
Remove showtitle attribute from asciidoc renderer

skybon:
Middleman-called Asciidoctor instance renders title no matter what. The "showtitle" attribute specified in options in asciidoc.rb is not overrideable. This is very annoying if you have custom header (title gets repeated twice: in your header and start of the body).

At the same time the rest of .asciidoc header is only stored in meta.

@mojavelinux I believe that such behaviour was not intended by design and this PR disables the offending attribute.

### DIFF
--- a/middleman-core/fixtures/asciidoc-app/source/hello-with-title.adoc
+++ b/middleman-core/fixtures/asciidoc-app/source/hello-with-title.adoc
@@ -1,4 +1,5 @@
 = Page Title
+:showtitle:
 :page-layout: default
 
 Hello, AsciiDoc!
--- a/middleman-core/lib/middleman-core/renderers/asciidoc.rb
+++ b/middleman-core/lib/middleman-core/renderers/asciidoc.rb
@@ -8,7 +8,7 @@ module Middleman
           app.config.define_setting :asciidoc, {
             safe: :safe,
             backend: :html5,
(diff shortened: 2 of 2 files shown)


mojavelinux:
Correct, this should be something the document can control. That either means it shouldn't be set or it should be set to `showtitle=@`. The latter allows the document to win. For now, we can take it out, since it's easy enough to add back in using the configuration.

tdreyno:
@mojavelinux Can you confirm that this PR is the correct solution?

mojavelinux:
@tdreyno There isn't really a right or wrong solution here. Certainly, this pull request enables the document to control whether the title is shown (using an `<h1>` element) in the embeddable content, so that's good. The fact that it's not enabled by default is a change, but perhaps not the wrong one. That should be communicated in the release notes, though.

In other words, the correct solution is a toss up between setting `showtitle=@` or not setting it at all. Either way allows the page to control whether it is included, which is a step forward.

For what it's worth, it is not enabled by default in the Jekyll plugin. See https://github.com/asciidoctor/jekyll-asciidoc/blob/master/lib/jekyll-asciidoc.rb#L20

Of course, the user can customize these attributes in the Middleman configuration, so whatever we decide can still be changed.

Issue #1561, opened 2015-07-15
Aruba 0.8 breaks the middleman build


mauro-oto:
Running the tests locally and in Travis, the build breaks when using Aruba 0.8. `bundle install` installs the latest version for it, released yesterday. 

In my PR https://github.com/middleman/middleman/pull/1560 I locked the Aruba gem to `~> 0.7.4` instead so that it doesn't update to 0.8 yet.

mauro-oto:
Closed after merge of https://github.com/middleman/middleman/pull/1560

maxmeyer:
@mauro-oto Can you tell me, what exactly breaks `middleman`? Are there any logs you can file as issue at http://github.com/cucumber/aruba?

mauro-oto:
@maxmeyer for some reason, on Mac OSX 10.10, it was not creating the directories it was supposed to create. I don't have many details as I didn't look into it too much, but I can reproduce this later and provide you some logs.

maxmeyer:
Ok. That should be fixed with 0.8.1...

PR #1562, opened 2015-07-15
Bump aruba

Arcovion:
Seeing if 0.8.1 works re: https://github.com/middleman/middleman/issues/1561#issuecomment-121627220

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'yard', '~> 0.8', require: false
 
 # Test tools
 gem 'pry', '~> 0.10', group: :development
-gem 'aruba', '~> 0.7.4'
+gem 'aruba', '~> 0.8.1'
 gem 'rspec', '~> 3.0'
 gem 'fivemat', '~> 1.3'
 gem 'cucumber', '~> 1.3'


maxmeyer:
Somehow `capybara` comes into scene, when the new matchers of `aruba` are used. Strange...

maxmeyer:
```
[1] pry(#<Object>)> expect(files).to all be_an_existing_file
ArgumentError: rack-test requires a rack application, but none was given
from /home/user/.gem/ruby/2.2.2/gems/capybara-2.4.4/lib/capybara/rack_test/driver.rb:16:in `initialize'
```

maxmeyer:
If you comment out the following line in `lib/middleman-core/step_definitions/server_steps.rb`:

``` ruby
# require 'capybara/cucumber'
```

And then run

```
bundle exec cucumber features/i18n_builder.feature:147
```

it works fine.

maxmeyer:
@jnicklas Any idea? Have you ever seen such a behaviour?

maxmeyer:
Got it.  `capybara` uses a method `rspec` also uses: `all` - see [RSpec's documentation](https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/all-matcher) for more on this matcher.

```ruby
  if negated
    expect(files).not_to include an_existing_file
  else
    expect(files).to all be_an_existing_file
  end
```
Looking at the source of `~/.gem/ruby/2.2.2/gems/capybara-2.4.4/lib/capybara/dsl.rb` there seem to be some lines of code relevant here:

```ruby
def page
  Capybara.current_session
end

Session::DSL_METHODS.each do |method|
  define_method method do |*args, &block|
    page.send method, *args, &block
  end
end
```

At `~/.gem/ruby/2.2.2/gems/capybara-2.4.4/lib/capybara/session.rb:24` you will find this:

```ruby
    NODE_METHODS = [
      :all #[...]
    ]
```

I'm not sure how we best handle this. Will create an issue over at capybara.

maxmeyer:
Opened an issue over at `capybara`. @tdreyno Ping.

maxmeyer:
This seems to be a known issue. I ask the guys at `rspec` for some help.

tdreyno:
So, this is not ready to merge, right?

Arcovion:
Right, and might not be so simple as just updating the gem version...

PR #1564, opened 2015-07-16
Removed most deprecation warnings from the build.

mauro-oto:
Hey, 

This PR removes most deprecation warnings from the build, as it's polluting the Travis log. They will be gone entirely after upgrading to `aruba ~> 0.8`, but this should take care of most of them.

Please check it out, thanks!


Arcovion:
Thanks

PR #1580, opened 2015-08-05
Bump version of cucumber

maxmeyer:
Let's try this one to get rid of https://github.com/middleman/middleman/pull/1528#issuecomment-128084209


maxmeyer:
Problem seems to be fixed? @tdreyno Or do you have any other problem with the build on your machine?

PR #1581, opened 2015-08-10
Fix misnamed config value which was causing --force-polling flag to b…

jsdalton:
…e ignored by file_watcher extension


jsdalton:
Note that in the preview server, here:

https://github.com/middleman/middleman/blob/master/middleman-core/lib/middleman-core/preview_server.rb#L110

The force polling config value is called `:watcher_force_polling` instead of `:force_polling`. It appears that these values have been out of alignment with each other since bedf235ff68998c0fd87f7531c860f82381312c2.

PR #1599, opened 2015-09-11
Ignore gems installed in local dir.

nslocum:
When installing gems in the local dir, the convention with Bundler is to install them in `/vendor/bundle`.
Ignoring that folder by default would be helpful to many.

### DIFF
--- a/middleman-core/lib/middleman-core/templates/shared/gitignore
+++ b/middleman-core/lib/middleman-core/templates/shared/gitignore
@@ -16,3 +16,6 @@
 
 # Ignore .DS_store file
 .DS_Store
+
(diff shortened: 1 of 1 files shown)


middleman-core/lib/middleman-core/templates/shared/gitignore
```diff
@@ -16,3 +16,6 @@
 
 # Ignore .DS_store file
 .DS_Store
+
+# Ignore gems installed in local directory
+vendor/bundle/
```
tdreyno: How about `/vendor/bundle/`? Vendor is a pretty common folder name and could be down in assets or node_modules paths.

PR #1604, opened 2015-09-19
Refactor frontmatter parsing, closes #1603

Arcovion:
Do not merge.
Doesn't work yet, idk why... ¯\\\_(ツ)_/¯
Will wait till the tests are working and then figure out where I went wrong.

PR #1608, opened 2015-09-23
Fix frontmatter parsing changes from #1608

Arcovion:
When I saw [this code](https://github.com/middleman/middleman/blob/409a892da707aab0eef848ae79fc7ad97700199d/middleman-core/lib/middleman-core/util/data.rb#L49-L53) checking for `coding: UTF-8` I thought it looked fragile and decided to just ignore everything before the frontmatter in #1604. I've added the check back in here as that didn't work.

### DIFF
--- a/middleman-core/lib/middleman-core/util/data.rb
+++ b/middleman-core/lib/middleman-core/util/data.rb
@@ -23,19 +23,22 @@ module Middleman::Util::Data
       return [{}, nil]
     end
 
-    case known_type
-    when :yaml
-      return [parse_yaml(content, full_path), nil]
-    when :json
-      return [parse_json(content, full_path), nil]
-    end
-
     /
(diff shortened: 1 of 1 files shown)


Arcovion:
Wow, this worked first time without even testing it: https://github.com/middleman/middleman/commit/450a333aa9d9766194fe0cf13856b17efbbfad8c
I've kept this as two commits, I think it covers everything in https://github.com/middleman/middleman/pull/1604#issuecomment-142694356

Arcovion:
Confirmed those three are fixed, the "A template should handle an empty YAML feed" one fails because of the unrelated tests regarding changing frontmatter which were already broken.

tdreyno:
Thanks. Yeah, fighting some aruba pathing issues which are placing file changes in the wrong place :(

PR #1611, opened 2015-09-30
Fix frontmatter regex

Arcovion:


Arcovion:
Was matching `--;` etc instead of `---`. Also the `encoding: UTF-8` line could have text above or below it, now it must be on it's own line directly above the opening frontmatter tag.
...

PR #1614, opened 2015-10-02
Little bit cleaned up. Dry in rake tasks. Fix $LOAD_PATH

poporul:


tdreyno:
Thanks. Let me review on Monday.

PR #1616, opened 2015-10-03
remove unnecessary require

y-yagi:
`require 'rubygems'` is already required in Ruby 1.9 or later.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,4 +1,3 @@
-require 'rubygems' unless defined?(Gem)
 require 'rake'
 
 require File.expand_path('../middleman-core/lib/middleman-core/version.rb', __FILE__)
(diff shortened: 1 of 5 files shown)


tdreyno:
Stable still supports old Ruby, but can you target this PR against the `master` branch which is Ruby 2.0+

y-yagi:
Thanks for reviewing. I will create another PR to master branch.

By the way, as long as  see the gemspec, stable branch appears to have only supports 1.9 or later.

https://github.com/middleman/middleman/blob/v3-stable/middleman-core/middleman-core.gemspec#L19
https://github.com/middleman/middleman/blob/v3-stable/middleman/middleman.gemspec#L19

do I misunderstand something?

tdreyno:
You may be right, I'd been a long time since we branched from stable.

That said, this build is failing on `1.9.3`: https://travis-ci.org/middleman/middleman/builds/83407233

tdreyno:
Hmm, that actually looks unrelated. Merging.

y-yagi:
Thanks!

Issue #1612, opened 2015-09-30
Custom renderer not receiving current_page


spiffytech:
I'm using a custom renderer that handles Mustache templates, but it's not receiving the per-page data (`current_page`). It's only receiving the `data` hashes (which you cant see I'm passing it in the `locals` hash key).

I cannot figure out how to pass the `current_page` data to my Mustache template, and cannot find any documentation on this. How can I do it?

```
# Require Gem
require 'tilt-mustache'

module Middleman
  module Renderers
    # Liquid Renderer
    module MiddlemanMustache
      # Setup extension
      class << self
        # Once registerd
        def registered(app)
          require 'mustache'

          app.before_configuration do
            template_extensions mustache: :html
          end

          # After config, setup liquid partial paths
          app.after_configuration do
            Mustache.template_path = source_dir

            # Convert data object into a hash for mustache
            sitemap.provides_metadata %r{\.mustache$} do
              { locals: { data: data.to_h } }
            end
          end
        end

        alias_method :included, :registered
      end
    end
  end
end
::Middleman::Extensions.register(:mustache, Middleman::Renderers::MiddlemanMustache)
```

spiffytech:
It appears the `current_page` data is showing up under `data.page`, but only *sometimes*. 

When shown from `middleman build`, the data is never present.

When run from `middleman server`, there is some sort of caching going on. This is what my workflow looks like (what a particular variable value is set to for a given page, and what shows up on the page. This applies to the entire page data structure - it's stale until a refresh. This happens with cURL too, so it's not my browser cache).

```
+-------------+     
|   (=true)   |     
|             |     
|    true     |     
+-----+-------+     
      |             
      | Page change 
      |             
+-----v-------+     
|  (=false)   |     
|             |     
|    true     |     
+-----+-------+     
      |             
      | Refresh page
      |             
+-----v-------+     
|  (=false)   |     
|             |     
|    false    |     
+-----+-------+     
      |             
      | Page change 
      |             
+-----v-------+     
|   (=true)   |     
|             |     
|    false    |     
+-----+-------+     
      |             
      | Refresh page
      |             
+-----v-------+     
|   (=true)   |     
|             |     
|    true     |     
+-------------+     
```

tdreyno:
In v3, `current_page` is a global value (which is attempting to lock itself for concurrent requests) that should be available on the `app`. I believe Mustache prefers to have concrete data (not instances) and avoid if statement, so I'm not sure how you'd get that information into the template.

[...]

PR #1617, opened 2015-10-03
Failing unit test for reading current_page values in Mustache

spiffytech:
In reference to issue #1612


tdreyno:
This is really great. I'll dive into how the mustache ruby library works and see what needs to get called to forward our environement.

PR #1626, opened 2015-10-07
Add support for Docker in development.

exegete46:
I wanted to use Middleman for a web designer friend, so I added Dockerfiles and documentation on how to use them.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,9 @@
 FROM ruby
 MAINTAINER Jack M <exegete@mac.com>
 
-RUN gem install execjs therubyracer
+RUN apt-get update \
+  && apt-get install -qqy nodejs \
+  && rm -rf /var/lib/apt/lists/*
 
 COPY . /src/
 RUN gem build /src/middleman/middleman.gemspec \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM ruby
+MAINTAINER Jack M <exegete@mac.com>
+
+RUN gem install execjs therubyracer
+
+COPY . /src/
+RUN gem build /src/middleman/middleman.gemspec \
+  && gem install ./middleman-*.gem \
+  && rm -Rf /src/
+
+VOLUME /project/
+WORKDIR /project/
+
+EXPOSE 4567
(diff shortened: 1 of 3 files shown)


tdreyno:
Is this for developing MM or for building a site in MM?

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ruby
+MAINTAINER Jack M <exegete@mac.com>
```
maxmeyer: Who is Jack M?

maxmeyer:
@tdreyno I don't see any reason to add such a thing to the `middleman` repo. It looks like he wants to bundle middleman and middleman-server and want to run it.

I suggest that you @exegete46 setup your own repository and just install `middleman` within that docker image. Maybe something similar to https://github.com/fedux-org/centos-geminabox/blob/master/files/latest/Dockerfile or just `gem install middleman -v 3.4.0`.

exegete46:
@tdreyno This would be for developing a site with Middleman.  I suppose it could be used for development work on MM as well, but I don't think I have the experience creating a gem to ensure it is done right.

@maxmeyer I decided to put in the effort after looking at Docker Hub where I found [17 repositories that appear to be MM](https://hub.docker.com/search/?q=middleman) (I've since added mind as well).  With how easy it seems to keep Docker images up to date, I figured it was worth offering the Dockerfiles back the project.

Dockerfile
exegete46: I am, but I have no issues changing that to an official maintainer.  It's just habit to put myself in.

tdreyno:
Seems like this belongs in the "new project template" as an option. https://github.com/middleman/middleman-templates-default

Similar to the Rackup option.

exegete46:
The `Dockerfile.dev` would be a great fit for the templates.  I can work on moving that file over there, if the other `Dockerfile` is useful in the main app?  Heck, a `docker-compose.yml` would be nice over there, too.  However, in order to use `middleman init` to create project, you'd want the docker image from the main project.

tdreyno:
I think it might be easiest to simply host a generic Dockerfile somewhere (like Gist) and link to it from the docs.

PR #1633, opened 2015-10-09
Refactor Data Loader to prevent middleman from crashing due to invalid…

dg-ratiodata:
… data file.

Use case:
I use `middleman` to generate some error pages for our proxies. Normally I use `middleman server` to have a "live" preview while I change the data file with the contents for all error messages. While I work on the error messages it can happend that they contain invalid YAML. Unfortunately that makes the preview server crash. I refactored the data loader a bit.

I use the same pattern here like I did in some components of the preview server. Would you mind to merge it to v4 yourself?

### DIFF
--- a/middleman-core/features/data.feature
+++ b/middleman-core/features/data.feature
@@ -51,3 +51,31 @@ Feature: Local Data API
     Then I should see "title1:Hello"
     Then I should see "title2:More"
     Then I should see "title3:Stuff"
+
+  Scenario: Invalid YAML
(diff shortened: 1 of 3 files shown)


poporul:
:+1:

tdreyno:
This just got refactored on master by @Arcovion, so I don't think it will apply cleanly. I do like this approach, though.

maxmeyer:
Thanks anyway. :-)

PR #1647, opened 2015-10-23
Frontmatter delims

Arcovion:
Ok so there is weird stuff going on here, testing with `bundle exec cucumber features/front-matter.feature `:

1. Changing `frontmatter_delims` to `Middleman::Application.config[:frontmatter_delims]` in `data.rb:27` still works, I'm pretty sure it didn't when I tried the other day...
2. The `config[:frontmatter_delims][:yaml]` variable is mutated with `config[:frontmatter_delims][:yaml] << %w(\\--- \\---)` in each step; `%w(\\--- \---)` is added every time (this does not affect the validity of the test).
3. It makes no difference if you change it to `config[:frontmatter_delims][:yaml] = [%w(--- ---), %w(--- ...), %w(\\--- \\---)]`, it still ends up being a mess like `{:json=>[[";;;", ";;;"]], :yaml=>[["---", "---"], ["---", "..."], ["\\---", "\\---"], ["\\---", "\\---"], ["\\---", "\\---"], ["\\---", "\\---"], ["\\---", "\\---"], ["\\---", "\\---"], ["\\---", "\\---"], ["\\---", "\\---"]]}`
4. Commenting out `config.rb` makes no difference, so `# config[:frontmatter_delims][:yaml] << %w(\\--- \\---)` still takes effect even though ruby should not evaluate this code. Removing the line entirely works though, so it's not some cache bug.
5. The `/(?<var>.*)/ =~ text` syntax stopped working when I interpolated the other regexes, making the code much uglier. I don't know if that's a bug in ruby or what...

I passed the config option in from the extension as mentioned in https://github.com/middleman/middleman/pull/1636#issuecomment-150448522, I used the test from https://github.com/DarylWright/middleman/commit/1ab5fcd9d406d5cd36580edc0fb07cf00611fbe5, will wait to see if this breaks other tests...

### DIFF
--- a/middleman-core/features/front-matter.feature
+++ b/middleman-core/features/front-matter.feature
@@ -28,13 +28,18 @@ Feature: YAML Front Matter
     Then I should not see "layout: false"
     Then I should not see "title: Pandoc likes trailing dots..."
 
+  Scenario: Rendering Haml (yaml)
+    Given the Server is running at "frontmatter-app"
+    When I go to "/front-matter-haml.html"
+    Then I should see "<h1>This is the title</h1>"
+    Then I should not see "---"
 
   Scenario: YAML not on first line, no encoding
     Given the Server is running at "frontmatter-app"
     When I go to "/front-matter-line-2.html"
     Then I should see "<h1></h1>"
     Then I should see "---"
-    
+
   Scenario: YAML not on first line, with encoding
     Given the Server is running at "frontmatter-app"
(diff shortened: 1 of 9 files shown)


tdreyno:
Going to merge this. Only broken test that looks suspicious is `Scenario: Build and clean an app under a hidden directory`, but I'll investigate that post-merge. Thanks @Arcovion

Issue #1645, opened 2015-10-20
It would be nice if Middleman could run as a daemon background process


NathanBowers:
For example, Webrick has the **-d** option and Rack has **--daemonize**.

That way I could leave Middleman running on a dev server or my local Mac without having to keep a terminal open.

Note that **$ middleman &** works, as in it keeps the process alive after closing terminal, but the web server is unresponsive.

maxmeyer:
What about running it with `screen`? Or write a `unit-file` if you're using `systemd` (dev server)? Did you tried that?

maxmeyer:
Maybe you need to install `screen` first. Alternatively use `tmux` what I prefer over `screen`.

johnnyshields:
This should be fixed now with https://github.com/middleman/middleman/pull/1683

NathanBowers:
Works! Awesome! :sparkling_heart:

allthesignals:
This is great! But I think it breaks livereload when overriding the host. 

```
configure :development do
  activate :livereload, host: "0.0.0.0"
end
```

`middleman serve --bind-address 0.0.0.0 -d`

logemann:
Can it be that when putting it in background via "-d" middleman no longer sees any file changes?

PR #1649, opened 2015-10-23
Support for --daemonize flag

twinn:
Daemonize the middleman process when the --daemonize flag is passed.

#1645


tdreyno:
Is this something desired on every subcommand or only the preview server? I'd suggest adding the flag to the `server` command, so it can be documented and have a short hand `-d` as well.

maxmeyer:
@twinn Are you working on this?

PR #1655, opened 2015-10-26
resolves #1653 don't rewrite paths that start with //

### LINKED ISSUE #1653: asset_host should not modify URLs that begin with //
As of Middleman 4.0.0.rc.1, asset_host modifies URLs that begin with `//`. The prefix `//` indicates an absolute URL and should not be modified. This affects the Google Analytics URL, among other examples.

mojavelinux:


mojavelinux:
I'm not sure if this is the right place to fix this issue, but it's the same place that excludes absolute URLs, so I figured it was the best place to start. This really stems from a problem in addressable because it considers the path `//example.com/path` to be relative (i.e., the call to `relative?` returns true). The way to detect this type of URL is to see that the scheme is nil but the host is not nil. Thus, a relative path can be properly detected if both of these properties are nil.

tdreyno:
Yeah, looks like this happened when we moved from a regex to a URI parser. Thanks!

mojavelinux:
Thanks @tdreyno!

PR #1657, opened 2015-10-26
resolves #1656 don't mangle URL inside tag

### LINKED ISSUE #1656: asset_host is mangling linked text that looks like a URL
The asset_host extension (as of Middleman 4.0.0.rc1) is mangling linked text that looks like a URL. Here's an example:

```
<a href="https://github.com/angular/angular.js">Angular.js</a>
```

Assuming the host is configured as `http://assets1.example.com`, here's the result we get:

```
<a href="https://github.com/angular/angular.js"http://assets1.example.com/>Angular.js</a>
```

mojavelinux:


mojavelinux:
I'm not sure if this is the best way to exclude this type of match. Frankly, the way URLs is currently being detected is very dangerous in my mind, so we're just whacking moles here. I'd much prefer to move to a much more strict set of known match locations and perhaps require the regex to be configured externally.

tdreyno:
Can you rebase please?

mojavelinux:
Certainly! Stay tuned.

mojavelinux:
Rebase completed @tdreyno.

PR #1659, opened 2015-10-27
Add link for RubyGem package manager download

glamouracademy:


tdreyno:
In what context is Rubygems not available?

glamouracademy:
This is just a link to the RubyGem package manager download that is mentioned in the Readme

tdreyno:
Right, but RubyGems is included with Ruby by default now, IIRC. So who would need a download link?

glamouracademy:
Ah yes I understand your point. I was thinking back to when I used Middleman. At that time I was more junior (than I am now). I really enjoyed the ease of use and accessibility of it, but sometimes even if libraries and add ons are included with Ruby (or even with Rails), I find it useful to research what it is. I believe this link doesn't harm anyone and if someone is wanting more information, it's an easy, one click to get it.

tdreyno:
Makes sense to me. Thanks @glamouracademy !

PR #1661, opened 2015-10-28
Update donation link

tysongach:
It appears that [Spacebox](http://spacebox.io) became [Plasso](https://plasso.co) and along the way, I think Spacebox’s certificate must've been retired or something. So visiting the current link shows a “Your connection is not private” error (at least, in Chrome).

This PR updates the README to point to the new Plasso link.


tdreyno:
Thanks1

PR #1668, opened 2015-10-29
Update link to getting started guide

djpowers:
Currently, the link to the guide goes through a redirect or two. This change uses the destination URL directly.


tdreyno:
Thank you

PR #1679, opened 2015-11-09
Upgrade to aruba 0.10.x [WIP]

dg-ratiodata:
@tdreyno @Arcovion 

1. I'm tired to change the `Gemfile` on each contribution. Since this is backward-compatible I changed it to download the git repository via HTTP to make it work across HTTP proxies
2. This upgrades your test suite to aruba 0.10.x and removes all methods with deprecation warnings. Since changes to ENV are "bundle" with aruba from 0.8.0 or so, I modified all places where ENV is accessed. I also make sure, that changes of CWD are local to the steps to prevent "leaking".

To see if this works across all of your supported platforms, I sent you this PR. Both changes are made in two separate commits.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,4 +37,4 @@ gem 'coveralls', '~> 0.8', require: false
 # Middleman itself
 gem 'middleman-core', path: 'middleman-core'
 gem 'middleman', path: 'middleman'
-gem 'middleman-sprockets', github: 'middleman/middleman-sprockets', branch: 'v3-stable-real'
+gem 'middleman-sprockets', git: 'https://github.com/middleman/middleman-sprockets', branch: 'v3-stable-real'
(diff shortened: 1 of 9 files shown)


tdreyno:
Looks pretty good. Only stylus failing, which sometimes results when there isn't a Node environment available. Possible side-effect of this PR?

dg-ratiodata:
Unfortunately not. We're now pretty strict with changing the environment. the stylus gem modifies PATH and that get's lost. I need to look for solution for this.

tdreyno:
Hate to ask, since it's my own damn fault, can this be easily ported to `master`?

maxmeyer:
Mmh... Should be. No problem, will have a look at it next week.

But @tdreyno "stylus" is still a problem and you've got another failing test. I need to fix this [PR](https://github.com/cucumber/aruba/pull/321), I created some days ago until this will work. Unfortunately I don't see a straight forward solution for this problem now - at least if deleting environment variables should work to.

tdreyno:
I don't really care about the stylus test

PR #1683, opened 2015-11-11
Make middleman's preview server a daemon

maxmeyer:
@tdreyno Closes #1649 and makes it obsolete. This makes middleman's preview server become a daemon.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ master
 * Rather than applying layouts to all files which are not .txt, .css, .js, .json: the new behavior is to only default layouts to active for .html
 * Switch from Ruby Sass to SassC.
 * `relative_assets` extension overrides local `relative: false` option to stylesheet/javascript tag helpers.
+* Add `before_server`-hook to the preview server which is run before the Webrick server is started
+* Add `-d` to `middleman server` to make it run as daemon
 
 # 4.0.0.rc.1
 
--- a/middleman-cli/lib/middleman-cli/server.rb
+++ b/middleman-cli/lib/middleman-cli/server.rb
@@ -49,6 +49,11 @@ module Middleman::Cli
                  aliases: '-l',
                  default: 0.5,
                  desc: 'Set file watcher latency, in seconds'
+    class_option :daemon,
+                 type: :boolean,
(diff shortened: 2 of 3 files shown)


tdreyno:
Mind adding a line to `CHANGELING`?

maxmeyer:
Fixed.

maxmeyer:
This is v4 only.

jimothyGator:
It seems `fsevent_watch` doesn't run when the `-d` option is used, so changes are not picked up and middleman must be restarted when files are changed. Is this intentional?

Tested on 4.0.0.rc.2, OS X 10.11.2, Ruby 2.2.3p173

tdreyno:
Not intentional, but I don't know enough about how `fork` and `Thread.new` work to figure our where it's going wrong.

maxmeyer:
@jimothyGator Not sure what went wrong here. Do you have the chance to check if it works on linux?

jimothyGator:
@maxmeyer I just tested on Ubuntu 14.04, and changes are not picked up there, either. If I run in daemon mode, I have to kill and restart the process for any changes to get picked up. I tested this with a new middleman project, making edits to `source/index.htm.erbl` and reloading the page.

I don't see a `fsevent_watch` process running on Ubuntu, whether I'm running in daemon mode or not. However, using `htop`, I see seven `middleman` processes running normally, versus only two running in daemon mode.

This may be a bit vague, but hopefully that provides some useful clues.

@NathanBowers You reported this is working for you on OS X. (#1645). As you make changes to your source files, are they reflected in the browser (either via LiveReload or manually reloading)? Perhaps I'm just doing something silly.

The command I run (on both OS X and Ubuntu) is `bundle exec middleman server -d`. Nothing special in my `config.rb`; as I mentioned, I also see this with a newly created middleman project.

tdreyno:
Uh, 7 detached MM processes? I may try killing those

NathanBowers:
@jimothyGator Doh! You're right. (I didn't bother to test changing the files with `-d`).

Changes are not picked up by daemon. For example, after I start with `-d` I can delete my layout files but then refresh the browser and it's as if the layout is still there.

PR #1694, opened 2015-11-27
Fix SASS live reload

johnnyshields:
In SourceWatcher, add a subroutine to find files which should be considered "also dirty" when a given file is touched.


johnnyshields:
Replaced by #1695

Issue #1627, opened 2015-10-07
SASS globbing appears broken in v4.0.0.rc1


lachlanmcdonald:
I've been porting one of our frameworks over to v4.0.0.rc1 (just to see how much we have to refactor) and identified a change in behaviour. In 3.4 you could glob multiple SASS files using:

```sass
@import "test/**/*";
```

And everything under the `test` folder would be imported without a hitch. On 4.0.0.rc1 the same code throws the following error:

```
Sass::SyntaxError: File to import not found or unreadable: test\**\*.
Load paths:
  \Users\lacmcdonald\Sites\mmtest\source\javascripts
  \Users\lacmcdonald\Sites\mmtest\source\stylesheets
```

To test, I ran `middleman init` and and moved:

- `stylesheets/site.css` to `stylesheets/test/_site.scss`
- `stylesheets/normalise.css` to `stylesheets/test/_normalise.scss`

Then created a new `site.css.scss` file with the above `@import`. I opted in to the Asset Pipeline and Compass during the init.

Might have just done something stupid during the init, but globbing is a pretty big win for us with some projects spread across hundreds of partials, so hope we can find a solution.

tdreyno:
@lachlanmcdonald Do you have `middleman-sprockets` in your `Gemfile.lock`?

Can you point your gems at `master`? This might have already been fixed.

lachlanmcdonald:
Here are the contents of the `Gemfile`:
```
source 'https://rubygems.org'
gem 'wdm', '~> 0.1.0', platforms: [:mswin, :mingw]
gem 'tzinfo-data', platforms: [:mswin, :mingw, :jruby]
gem 'middleman-sprockets', '>= 4.0.0.rc.1'
gem 'middleman-compass', '>= 4.0.0.rc.1'
gem 'middleman', '>= 4.0.0.rc.1'
gem 'middleman-livereload'
```

tdreyno:
@lachlanmcdonald If you remove `middleman-compass`, do you have the same issues?

lachlanmcdonald:
@tdreyno Same error persists.

Here's a copy of the full output, `Gemfile` and `Gemfile.lock`:
https://gist.github.com/lachlanmcdonald/40a2c688e5c34dd9f692

tdreyno:
Great, thank you. Will resolve before next RC

johnnyshields:
PR to fix this issue is here: https://github.com/middleman/middleman/pull/1695

[...]

PR #1695, opened 2015-11-27
Fix SASS partial live reloading

### LINKED ISSUE #1631: Stylesheets not updating in middleman v4
Stylesheet partials aren't being updated when making changes.
If I change them to normal stylesheets (remove the underscore), everything works well

johnnyshields:
Fixes #1631, #1627

In SourceWatcher, add a subroutine to find files which should be considered "also dirty" when a given file is touched.

### DIFF
--- a/middleman-core/lib/middleman-core/sources/source_watcher.rb
+++ b/middleman-core/lib/middleman-core/sources/source_watcher.rb
@@ -222,6 +222,8 @@ module Middleman
 
       return if updated.empty? && removed.empty?
 
+      updated |= find_related_files(updated + removed)
+
       update(updated.map { |s| Pathname(s) }, removed.map { |s| Pathname(s) })
     end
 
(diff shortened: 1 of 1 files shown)


middleman-core/lib/middleman-core/sources/source_watcher.rb
```diff
@@ -222,6 +222,8 @@ def on_listener_change(modified, added, removed)
 
       return if updated.empty? && removed.empty?
 
+      updated |= find_related_files(updated + removed)
```
tdreyno: Should this be `||=`

tdreyno:
Hi @johnnyshields, thanks for debugging this. I'm going to try to take a slightly more holistic approach that handles all partials, not just Sass. I'll let you know in a few if it makes any sense

tdreyno:
Actually, merging this as a great starting place.

tdreyno:
@johnnyshields Can you verify https://github.com/middleman/middleman/commit/4594f46fb8ca8227e50e51a9706224cdd1ea5397 didn't break your fix?

johnnyshields:
@tdreyno I can't run master as Ruby-SASSC is still broken. It looks like it should work tho.

johnnyshields:
Spoke too soon, you fixed it! Thanks.

johnnyshields:
@tdreyno your commit is broken:
1. `File.extname` includes the period, so you need to use this glob:  "**/[^_]*#{File.extname(file)}"
2. Even after fixing #1, I can't get SASS partials to load (please investigate).
3. Not such a big deal but did you consider that SCSS and SASS can be embedded inside each other? Also future versions of SASS will allow embedding CSS.
4. Note that SASS can embed files even if _ is not used. I tested this on your branch and it works, though I'm not sure why it works exactly--probably the partial hack is not needed for that case.

tdreyno:
Test-only PRs are very welcome and can always help get your specific use-case implemented.

Try now, rewrote with even larger scope and some basic RSpecs.

equinusocio:
Any update???

PR #1707, opened 2015-12-17
Skip front matter parsing for ignored resources

jordanandree:
I believe this is an accurate assumption to make for a resource – that if it is ignored then it shouldn't be parsed by the front matter extension.

Anything I'm missing on this? Should an resource in an ignored path need to have parsed front matter?


tdreyno:
Seems reasonable. Is this a performance concern or working around some other issue?

jordanandree:
Sounds good.
It's definitely just working around an issue, haven't done any perf testing
for it.

tdreyno:
What was the issue? Just making sure this is actually fixing the root cause.

jordanandree:
My issue is that I have an ignored directory that raises a front matter
exception when a file in it is parsed by the front matter extension. This
PR fixes it.

PR #1710, opened 2015-12-23
Exit with error code on `init` command failed

tricknotes:


maxmeyer:
Why not use a separate exit code for this? @tdreyno Just returning 1 is not that meaningful. Other applications return specific error codes on different errors.

tdreyno:
I was under the impression that `1` was the standard error exit code?

maxmeyer:
> I was under the impression that 1 was the standard error exit code?

Yes, personally I would describe it as the "CatchAll"-error code. Nothing you can react upon specifically on commandline, besides "Oh, an error occured".

Some applications exit with different exit codes depending on the error which caused the "abort". 

Some examples:
* https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html
* https://www.freebsd.org/cgi/man.cgi?query=sysexits&sektion=3
* http://tldp.org/LDP/abs/html/exitcodes.html

Books:
* Build Awesome Command-Line Applications in Ruby 2 (Look for "Play nicely with others - Exit Codes or Are there any default exit codes) on Amazon

Some more reading (for those with more time):
* https://practicingruby.com/articles/building-unix-style-command-line-applications

tdreyno:
Given that this is primarily for CI-servers, I imagine, 1 is fine. If you have a recommendation, I'd gladly accept that too. I simply don't know better.

maxmeyer:
Ok. Then let's choose 1 for now. :-)

PR #1711, opened 2015-12-23
Test against 2.2.4 that is the latest of Ruby 2.2

tricknotes:


tricknotes:
Oops, test failed. :broken_heart: 
I'll check it later :<

tdreyno:
It's a silly timing-related bug that I haven't debugged on Travis yet

Issue #1709, opened 2015-12-23
v4: i18n parent relations set wrongly when mixing localized and non-localized pages


GhostGambler:
The following little Middleman project:
``` shell
mkdir bugreport && cd bugreport
echo "gem 'middleman', '>= 4.0.0.rc.2'" > Gemfile
bundle install
bundle binstub middleman-cli
echo "activate :i18n, langs: [:de, :en]" >> config.rb
mkdir -p source/localizable/sub
mkdir -p source/sub
cat <<EOT > source/localizable/sub/test.html.erb
---
title: sub/test
---
<p>Page: sub/test</p>
<p>Path: <%=
hierarchy = [current_page]
hierarchy.unshift hierarchy.first.parent while hierarchy.first.parent
hierarchy.collect {|page| page.data.title }.join(" # ")
%></p>
EOT
cat <<EOT > source/sub/index.html.erb
---
title: sub/index
---
sub-index
EOT
./bin/middleman server preview
```

Consists of two callable urls with the following output:
```
http://localhost:4567/sub/test.html
Page: sub/test
Path: sub/test
```
```
http://localhost:4567/en/sub/test.html
Page: sub/test

Path: sub/index # sub/test
```
which is unexpected. In both cases should the page sub/index be part of the hierarchy.

See #1701 for the origin of the problem.

tdreyno:
Great. Seems to be easy to write a test for this.

GhostGambler:
I've written tests, which cover multiple problematic cases (which all do not work). I doubt I can solve the issue cleanly in reasonable time, so I am afraid the rest is up to you.

tdreyno:
Thanks again

PR #1733, opened 2016-01-05
Test for issue 1709; parent relations with localized and non-localized sources

GhostGambler:
Some tests for issue #1709


tdreyno:
Thanks!

tdreyno:
Wait, so the idea is that i18n paths terminate at the mount point, rather that the site root `/`? I think that makes sense.

Issue #1702, opened 2015-12-10
v4: non-utf8 characters prohibit execution


GhostGambler:
If a non-utf8 character is somewhere in the file, the regex, which searches for a frontmatter, fails:
```
/.../gems/middleman-core-4.0.0.rc.2/lib/middleman-core/util/data.rb:40:in `match': invalid byte sequence in UTF-8 (ArgumentError)
	from /.../gems/middleman-core-4.0.0.rc.2/lib/middleman-core/util/data.rb:40:in `parse'
	from /.../gems/middleman-core-4.0.0.rc.2/lib/middleman-core/core_extensions/front_matter.rb:68:in `data'
	from /.../gems/middleman-core-4.0.0.rc.2/lib/middleman-core/core_extensions/front_matter.rb:33:in `block in manipulate_resource_list'
	from /.../gems/middleman-core-4.0.0.rc.2/lib/middleman-core/core_extensions/front_matter.rb:30:in `each'
	from /.../gems/middleman-core-4.0.0.rc.2/lib/middleman-core/core_extensions/front_matter.rb:30:in `manipulate_resource_list'
	from /.../gems/middleman-core-4.0.0.rc.2/lib/middleman-core/sitemap/store.rb:206:in `block (2 levels) in ensure_resource_list_updated!'
	from /.../gems/hamster-2.0.0/lib/hamster/vector.rb:1316:in `each'
	from /.../gems/hamster-2.0.0/lib/hamster/vector.rb:1316:in `traverse_depth_first'
	from /.../gems/hamster-2.0.0/lib/hamster/vector.rb:431:in `each'
	from /.../gems/middleman-core-4.0.0.rc.2/lib/middleman-core/sitemap/store.rb:204:in `block in ensure_resource_list_updated!'
	from /Users/ulrichgabor/.rvm/rubies/ruby-2.1.1/lib/ruby/2.1.0/monitor.rb:211:in `mon_synchronize'
	from /.../gems/middleman-core-4.0.0.rc.2/lib/middleman-core/sitemap/store.rb:196:in `ensure_resource_list_updated!'
	from /.../gems/middleman-core-4.0.0.rc.2/lib/middleman-core/sitemap/store.rb:153:in `block in resources'
	from /Users/ulrichgabor/.rvm/rubies/ruby-2.1.1/lib/ruby/2.1.0/monitor.rb:211:in `mon_synchronize'
[...]

PR #1734, opened 2016-01-05
Issue #1702

GhostGambler:
Tests for issue #1702 and first fix to use the option again.

Still missing correct charset content-type headers for the preview server. (Maybe also for txt files and similar).


tdreyno:
Thank you

PR #1737, opened 2016-01-07
Add option to gzip all file extensions.

johnnyshields:


johnnyshields:
I guess not needed -- https://github.com/middleman/middleman/pull/1738 is a better solution.

Issue #1563, opened 2015-07-15
Middleman hangs on minifed json with asset_hash enabled


ryanmark:
I have this json file in my project: http://bl.ocks.org/mbostock/raw/4090846/us.json

When I run build, it hangs on this file. One time we just let it run and it eventually completed after an hour or so.

The snag is here:
https://github.com/middleman/middleman/blob/v3-stable/middleman-core/lib/middleman-more/extensions/asset_hash.rb#L94

I can recreate this issue with these few lines of ruby:

```ruby
re = /([=\'\"\(,]\s*)([^\s\'\"\)]+(\.woff2|\.woff|\.webp|\.ttf|\.svgz|\.svg|\.png|\.otf|\.js|\.jpg|\.jpeg|\.gif|\.eot|\.css))/
open('us.json').read.gsub(re) { |match| puts match }
```

If I pretty format the json before running it through this regex, it goes super fast.

The work around is simple, but not obvious. Smells like a bug to me.

tdreyno:
Interesting. Any difference on different ruby versions?

tdreyno:
I guess in the short term. Pretty print and re-minify at build time... or, add to `asset_hash`'s `ignore` array of regexes.

Arcovion:
You could introduce a max filename limit, which I think is 255:
```ruby
re = /([=\'\"\(,]\s*)([^\s\'\"\)]{1,255}(\.woff2|\.woff|\.webp|\.ttf|\.svgz|\.svg|\.png|\.otf|\.js|\.jpg|\.jpeg|\.gif|\.eot|\.css))/
```
Still bad, but won't take *hours* any more :D

tdreyno:
If I remove the captures, it's a lot faster to test for a single match before trying to replace all.

tdreyno:
Pushed a tweak. Filename limit plus the removal of an unneeded capture.

Arcovion:
I think the capture was needed to group the extensions, otherwise it's like `filename.foo|.bar` instead of `filename(.foo|.bar)`
Changing this line:
```ruby
 @exts_regex_text = @exts.map { |e| Regexp.escape(e) }.sort.reverse.join('|')
```
to
```ruby
 @exts_regex_text = Regexp.union(@exts.sort.reverse)
```
will fix it though, making a non-capturing group `(?:)` which looks like `/filename(?-mix:\.foo|\.bar)/`

ryanmark:
Thanks for looking into this so quickly!

[...]

PR #1740, opened 2016-01-08
Add rewrite_ignore option to asset_hash extension

GUI:
This allows for configuring the files that will be skipped when performing URL rewriting on the file content. This differs from the existing `ignore` option, which configures the files that won't have hashes added to the filename.

This stemmed from wanting to improve build times on a project by skipping the URL rewriting phase on some large minified JS files didn't contain any asset URLs. The discussion in https://github.com/middleman/middleman/issues/1563#issuecomment-122045822 also points to another potential use for this (skipping problematic files).

I did discover that it's currently possible to do something similar with a custom middleware, like this:

```ruby
class InlineURLRewriterIgnores
  def initialize(app, options= {})
    @app = app
  end

  def call(env)
    if(env["PATH_INFO"] =~ %r{^/example})
      env["bypass_inline_url_rewriter"] = "true"
    end

    @app.call(env)
  end
end
use InlineURLRewriterIgnores
```

However, this seems to use an undocumented option (`bypass_inline_url_rewriter`), and it's a little more cumbersome, so I thought I'd submit this pull request, which would make that same configuration look more like:

```
activate :asset_hash, :rewrite_ignore => [
  %r{^/example},
] 
```

This only adds the new `:rewrite_ignore` option to the `AssetHash` extension, but it would be easy to add this same handling to the other extensions that use the `InlineURLRewriter` middleware (`AssetHost`, `CacheBuster`, and `RelativeAssets`) if you'd be interested.

Let me know if you have any feedback on this approach, the name of the option, etc. Thanks!

### DIFF
--- a/middleman-core/features/asset_hash.feature
+++ b/middleman-core/features/asset_hash.feature
@@ -203,6 +203,45 @@ Feature: Assets get file hashes appended to them and references to them are upda
       | javascripts/application-1d8d5276.js |
       | stylesheets/site-7474cadd.css |
 
+  Scenario: Hashed-asset files are not replaced for rewrite ignored paths
+    Given a fixture app "asset-hash-app"
+    And a file named "config.rb" with:
+      """
+      is_stylesheet = proc { |path| path.start_with? '/stylesheets' }
+      activate :asset_hash, rewrite_ignore: [
+        %r(javascripts/*),
+        '/subdir/*',
+        is_stylesheet
+      ]
+      activate :relative_assets
+      activate :directory_indexes
+      """
+    And a successfully built app at "asset-hash-app"
+    When I cd to "build"
+    Then the following files should exist:
+      | index.html |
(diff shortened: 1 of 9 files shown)


tdreyno:
This is really smart. I do think it should be on all those similar file-parsing middlewares you mention.

Thanks. Can you update this PR with those other ones too? That'd be amazing.

GUI:
Sounds good, thanks for the feedback! And yup, I can get this pull request updated with the other extensions.

GUI:
@tdreyno: I've pushed an extra commit that adds the `rewrite_ignore` option to the all the other extensions using the `InlineURLRewriter` middleware.

And it looks like Travis is failing, but the same tests are failing on master, so I don't think it's related to these additions. But let me know if you spot any issues.

tdreyno:
So good. Thank you.

PR #1747, opened 2016-01-11
Add app method to detect mode

jsdalton:
Not sure if operating outside the two official "modes" of either `:build` or `:server` is supported, but if so this additional method would be convenient.

An example is the Contentful Middleman library (https://github.com/contentful/contentful_middleman), which for V3 had the concept of a `:contentful` environment when the task that pulls content down was run, but has now switched to setting `:contentful` as the mode for V4.

Anyhow, this method is just a convenience but I expected it's existence based on the parallel existence of the `environment?` method.

### DIFF
--- a/middleman-core/lib/middleman-core/application.rb
+++ b/middleman-core/lib/middleman-core/application.rb
@@ -327,18 +327,26 @@ module Middleman
       end
     end
 
+    # Whether we're in a specific mode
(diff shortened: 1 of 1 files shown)


tdreyno:
Yup, I made a `:contentful` mode in my own fork of their repo. This is a nice convenience method.

PR #1763, opened 2016-01-18
Fix gzip extension status output

iiska:
Proposed fix for #1364 

- Changed event name to :created when reporting created output files
- Added empty string as a second parameter to assign total savings summary as extra param which is printed on [Middleman::Cli::Build#on_event](https://github.com/middleman/middleman/blob/d82ac590db5fbb4f7482aed679acef9646d1c45c/middleman-cli/lib/middleman-cli/build.rb#L107) when event type is :gzip


tdreyno:
Thank you

PR #1781, opened 2016-01-27
Support for .xhtml

FearGoidte:
I’ve attempted to make support for .htm and .xhtml consistent.
Extensions are in alphabetical order to make it easier to compare them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ master
 * Correctly show file names of GZIP'ed assets. #1364
 * Build file output is not parallel-ized! Use `middleman build --no-parallel` to disable.
 * Make template file extensions that get layouts by default configurable via `config[:extensions_with_layout]`
+* Remove `=` from inline url matcher. This means paths in HTML attributes MUST be quoted. Fixes #1780
 
 # 4.0.0
 
--- a/middleman-core/features/asset_host.feature
+++ b/middleman-core/features/asset_host.feature
@@ -17,6 +17,8 @@ Feature: Alternate between multiple asset hosts
     When I go to "/stylesheets/asset_host.css"
     Then I should see content matching %r{http://assets1.example.com/}
     Then I should not see content matching %r{http://assets1.example.com//}
+    When I go to "/javascripts/asset_host.js"
+    Then I should not see content matching %r{http://assets1.example.com/}
 
   Scenario: Set proc host with inline-option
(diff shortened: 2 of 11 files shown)


tdreyno:
Looks good. Though... not entirely sure about automatically gzipping XML. And resources where I can read about how common that is?

FearGoidte:
I can’t think of a reason why <abbr title="eXtensible Markup Language">X.M.L.</abbr> would be unsuitable to Gzip rather than any other mark-up. With regard to regard to how common it is, for Gzipping <abbr title="eXtensible Markup Language">X.M.L.</abbr>: [Gzip Compressed XML Sitemaps: sitemap.xml vs sitemap.xml.gz](http://www.microsystools.com/products/sitemap-generator/help/xml-sitemaps-compression-gzip/). That restates what I would have guessed, searches for setting up apache for Gzip include plenty of examples that add <abbr title="eXtensible Markup Language">X.M.L.</abbr>. I know that minifying <abbr title="eXtensible Markup Language">X.M.L.</abbr> can break <abbr title="eXtensible Markup Language">X.M.L.</abbr> processing instructions but that changes the content.

If you are asking how many people are *using Middleman* and serving it as `application/xhtml+xml`; I would guess that I’m the only person doing that. Bear in mind that, my as yet unpublished musings are going to be about how to install Middleman and serve it as `.xhtml`, plus the odd post about <abbr title="Cascading Style Sheets">C.S.S.</abbr> or <abbr title="Resource Description Framework">R.D.F.a.</abbr> and most of it in Irish so if these changes have any impact on the performance of serving `.html` than they should obviously be ignored.

tdreyno:
Yeah, I guess I'm saying, adding `xhtml` is a benefit for anyone using that and there's no danger of backwards incompatibilities.

Adding `.xml` to the default gzip list, without knowing all the places where that xml is being consumed, "might" cause regressions. So I think I'd avoid that for now. Users can always opt-in to that when activating `gzip`.

FearGoidte:
I can’t see how Gzip might break `.xml` but as you say it’s easy enough to turn on in the settings. It seems to me it should be on by default but I have a long history of insufficient testing to undermine my opinions.

tdreyno:
Cool, mind updating the PR to remove `xml` in gzip and then I'll merge.

FearGoidte:
Updates made

tdreyno:
Thank you @FearGoidte

PR #1782, opened 2016-01-28
Disable file watching completely when watcher_disable is true

jsdalton:
Right now, when `watcher_disable` is true, Middleman doesn't actually disable the watcher, it merely adds a few additional ignore paths. Not sure if there's a reason for this? It seems like not the desired behavior when one sets watcher disable to true.

It'd be great if it *actually* disabled the watcher entirely.

This is sort of a nuclear option, but the advantage of this is that I can implement my own custom watch behavior separately, under my own ready callback.


tdreyno:
Seems reasonable, not sure why the if statement was scoped like tat.

PR #1794, opened 2016-02-05
Allow files from import_file or import_path to be rendered

stevenosloan:
@tdreyno this is by no means ready to merge, just wanted to get guidance on it early.

For back story, this is driven from an asset gem I maintain -- it contains a little glue to make it simple to mount in middleman, rails, plain sass or sprockets projects. Part of this involved mounting a few pages out of the gem for the preview server to have as reference for patterns. It used something very similar to [this](https://github.com/stevenosloan/middleman_external_resource_test/blob/master/lib/three_four_extension.rb) for 3.x (though now that I've looked at the implementation of `FileWatcher#reload_path` has the same caveat I'll touch on later).

In adding support for 4.x, I thought the new `import_file` or `import_path` would be a perfect fit and I could start using official APIs for things. They have some major caveats though, in the case of 4.0 a file can't have frontmatter, and in current 4.1.x the file can't be rendered at all (since it's set to type binary, and that's where a difference between `import_file` & `import_path` was introduced).

So this is an attempt at:

1. letting files loaded from `import_file` or `import_path` to be rendered & use frontmatter
2. have files loaded from `import_file` & `import_path` behave similarly
3. profit?

First thing I've done is consolidated the creation of sitemap resources across the two methods and used the `Util.binary?` method to determine if `:binary` should be added as a type -- and I feel pretty good about that.

What I don't feel good about is doing [this](https://github.com/middleman/middleman/commit/9c72b01a4f666ccf6b3f99d9ce0961adf438af0e#diff-c2956fcd887ca66c85532e4a272057d6R65) to get the files into a source watcher so frontmatter can be processed. In these minor test cases it works out *ok*, but as you've pointed out before if an imported file happened to be at the root of a deep directory adding that path to the watcher could cause performance problems. I have a few ideas that could work to help with that, but they could have fairly wide implications so direction would be great.

1. Create a "single file" source watcher -- I feel a little odd at creating a new watcher to add to the array that needs to be searched, as well as re-implementing the methods the source_watcher needs to respond to.

2. Add a layer above `app.files.find` that would look for files in source_watcher, then fall back to seeing if the file exists -- and only return nil if the file doesn't exists. Then update everywhere that calls this looking for file data (file_renderer is the main bottleneck I think). This comes with the caveat that template data wouldn't be cached for those files -- but that's probably an acceptable tradeoff, the often called files like partials/layouts should be in source and watched.

3. Punt on getting frontmatter working for files added this way in middleman proper -- and just implement the watcher hack in my project because I understand the implications and they don't affect my specific use case. ;)

So, that's a ton of words -- but I look forward to hearing your thoughts on this.


tdreyno:
Does `files.watch` not accomplish all of this?

stevenosloan:
It does in 4.0, but not 4.1 for files pulled with `import_file` (because `:binary` is always set as the type in 4.1 -- updated [this example project](https://github.com/stevenosloan/middleman_external_resource_test) with build output). With that one fix it'd work ok for me -- but if feels *super* unclear to know to do this, and what do you do if you want to import a renderable file from a directory with a ton of files?

I will admit it maybe feels like more of a problem to me because of how long it took puzzling over that behavior and tracking through how files in the sitemap are different that what renderers see. It was really surprising to have resources in the sitemap with a source_file that exists -- but can't get data from those files.

tdreyno:
I don't think you need to import if you're watching, unless you're doing a transformation? Or, as you say, only using a subset of the files. `watch` takes an `only` param which is an array of regexes to slim down the contents of the watch.

import is optimized for static assets, it was being used mostly for JS imports which would never have front matter and they were causing performance issues when reading every js file in a large project for front matter. So, they are considered "binary", which really means "static".

I think we may be talking about slightly different things here and I do think there is an omission and over optimization on the MM side. Can we back up and figure out what the actual feature is. Is it "store templates in a gem" kind of like a rails engine?

stevenosloan:
Oh right, I forgot that watching will pull the files into the sitemap as well -- so to do this I could watch w/ an only param for the file I want, then use a sitemap_manipulator (or move file maybe?) to get the file built where I want it to go.

To back up to the original goal, yeah -- render a file from outside of `source` into my project. (in my case it's in a gem, but could be a bower package or something as well).

tdreyno:
Yeah, I think `watch` to add an arbitrary folder to your site, `only` to limit the files and `move_file` to rename it if necessary.

stevenosloan:
Yeah, that gets me 85% of the way there. Color me overly optimistic that `import_file` was going to do exactly what I wanted ;)

I just happened to have a spec for this, but with a directory structure:

```
import/
  - index.html.erb
source/
  - index.html.erb
```

If I add a watcher for `import` it will clobber the index.html from source (of course if they didn't happen to have the same extensions it's all fine) -- because their relative paths will be the same, even if I wanted the file from import to end up somewhere else. 

Makes sense why that's happening -- since they're added as `:source` files they're all be expected to be uniq based on relative path. I can definitely work around that for my case, but I'm now realizing this is a source of some confusion I had around the external pipeline as well (didn't realize that was based around a file watcher). At some point I had created a filename collision because I was expecting to move files later and got pretty confused when one of them "disappeared" on me.

stevenosloan:
(going ahead and closing this because, as a PR, it's closed in my mind)

tdreyno:
Ah, gotcha. Yeah, watch (and thus pipeline) simply merge multiple trees of files.

You could have a `to_be_imported` folder to isolate names, then `move_file` them later. Thats "kinda" how i18n works.

stevenosloan:
Cool, yeah that's pretty much how I adjusted things. Thanks for talking that through, I got lost in internals and that got me stuck on the idea things needed to change.

PR #1805, opened 2016-02-15
Escape path in 404 (not found) page

melvinsh:
This escapes the `path` variable using `escape_html` from `Rack::Utils` to prevent (reflected) [Cross-Site Scripting](https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)) in 404 (Not Found) pages.

PoC: `http://localhost:9292/123<strong>HTML injected</strong>`.

screenshot 2016-02-15 23 40 37

`escape_html` is a good way to escape HTML from input. I'm not sure what the performance impact of including `Rack::Utils` is, if it even matters. I'm open to suggestions!

I wrote a test but I have no clue where it should go. I verified it though and it passes.

```
Scenario: Escapes HTML when a resource is not found
  Given the Server is running at "asset-hash-app"
  When I go to "/<p>HTML</p>"
  Then I should see "Not Found"
  And I should see "&#x2F;&lt;p&gt;HTML&lt;&#x2F;p&gt;"
  And I should not see "<p>HTML</p>"
```

### DIFF
--- a/middleman-core/lib/middleman-core/rack.rb
+++ b/middleman-core/lib/middleman-core/rack.rb
@@ -2,6 +2,7 @@ require 'rack'
 require 'rack/file'
 require 'rack/lint'
 require 'rack/head'
+require 'rack/utils'
 
 require 'middleman-core/util'
 require 'middleman-core/logger'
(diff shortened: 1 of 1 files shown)


melvinsh:
This obviously doesn't matter when you use middleman like you're supposed to use it. But it's likely that people are using rack to serve middleman sites (I personally discovered this when I was experimenting with Dokku and pushed a middleman project).

tdreyno:
Thanks

PR #1809, opened 2016-02-18
Make build fail when external_pipeline fails

adam-h:
This is to fix an issue where the build continues successfully when the external pipeline fails


tdreyno:
Excellent, thank you.

adam-h:
:+1: Thanks!

PR #1818, opened 2016-03-01
Remove compass importonce use fast blank

mauro-oto:
This PR uses the faster `blank?` from `fast_blank` instead of ActiveSupport's `blank?`, as the gem was already included but not being used. I also removed `compass-import-once`, which was not being used at all. 

Please check it out, thanks!


tdreyno:
I believe `import-once` is being used inside compass. It's auto-required by Bundler and does some caching magic

tdreyno:
I think fast_blank is also auto-required by Bundler.

PR #1826, opened 2016-03-03
Ensure that misspelled or missing partials generate an error

gltarsa:
Removed what looks like a stray debugging line from the code and added a test.


tdreyno:
Good catch, that does look suspicious.

PR #1829, opened 2016-03-08
Support CLI options as options

balthisar:
Supports visibility of the cli arguments in the application's config hash, as these
should/could be considered valid configuration options specified at the time of
application instantiation. The are exposed as config[:cli_options][:thor_key].
Note that this change _does not_ do any of this:
- does not allow the cli to override `config.rb` settings;
- does not define new cli/thor options.

It _does_ allow:
- Extensions can access cli options that they create when overriding Middleman::Cli
  classes, e.g., in `before_build` or `after_configuration`.

### DIFF
--- a/middleman-cli/lib/middleman-cli/build.rb
+++ b/middleman-cli/lib/middleman-cli/build.rb
@@ -51,6 +51,7 @@ module Middleman::Cli
       env = options['environment'].to_sym
       verbose = options['verbose'] ? 0 : 1
       instrument = options['instrument']
+      cli_options = options.to_hash.symbolize_keys
 
       builder = nil
 
@@ -61,6 +62,7 @@ module Middleman::Cli
           config[:mode] = :build
           config[:environment] = env
           config[:show_exceptions] = false
+          config[:cli_options] = cli_options
         end
 
         builder = Middleman::Builder.new(@app,
--- a/middleman-cli/lib/middleman-cli/build.rb
+++ b/middleman-cli/lib/middleman-cli/build.rb
@@ -36,12 +36,6 @@ module Middleman::Cli
                  default: false,
                  desc: 'Generate profiling report for the build'
 
-    # Allow extensions to verify cli arugments
-    # @return [void]
-    def check_cli_arguments
-      nil
-    end
-
     # Core build Thor command
     # @return [void]
     def build
--- a/middleman-cli/lib/middleman-cli/build.rb
+++ b/middleman-cli/lib/middleman-cli/build.rb
@@ -51,7 +51,7 @@ module Middleman::Cli
       env = options['environment'].to_sym
       verbose = options['verbose'] ? 0 : 1
       instrument = options['instrument']
-      cli_options = options.symbolize_keys.to_hash
+      cli_options = options || options.symbolize_keys.to_hash
 
       builder = nil
 
--- a/middleman-cli/lib/middleman-cli/build.rb
+++ b/middleman-cli/lib/middleman-cli/build.rb
@@ -51,7 +51,7 @@ module Middleman::Cli
       env = options['environment'].to_sym
       verbose = options['verbose'] ? 0 : 1
       instrument = options['instrument']
-      cli_options = options.to_hash.symbolize_keys
+      cli_options = options.symbolize_keys.to_hash
 
       builder = nil
 
--- a/middleman-cli/lib/middleman-cli/build.rb
+++ b/middleman-cli/lib/middleman-cli/build.rb
@@ -36,6 +36,12 @@ module Middleman::Cli
                  default: false,
                  desc: 'Generate profiling report for the build'
 
+    # Allow extensions to verify cli arugments
+    # @return [void]
+    def check_cli_arguments
+      nil
+    end
+
     # Core build Thor command
     # @return [void]
     def build
--- a/middleman-cli/lib/middleman-cli/server.rb
+++ b/middleman-cli/lib/middleman-cli/server.rb
@@ -55,6 +55,12 @@ module Middleman::Cli
                  default: false,
                  desc: 'Daemonize preview server'
 
+    # Allow extensions to verify cli arugments
+    # @return [void]
+    def check_cli_arguments
+      nil
+    end
+
     # Start the server
     def server
       require 'middleman-core'
--- a/middleman-cli/lib/middleman-cli/server.rb
+++ b/middleman-cli/lib/middleman-cli/server.rb
@@ -55,12 +55,6 @@ module Middleman::Cli
                  default: false,
                  desc: 'Daemonize preview server'
 
-    # Allow extensions to verify cli arugments
-    # @return [void]
-    def check_cli_arguments
-      nil
-    end
-
     # Start the server
     def server
       require 'middleman-core'
--- a/middleman-cli/lib/middleman-cli/server.rb
+++ b/middleman-cli/lib/middleman-cli/server.rb
@@ -79,7 +79,8 @@ module Middleman::Cli
         reload_paths: options['reload_paths'],
         force_polling: options['force_polling'],
         latency: options['latency'],
-        daemon: options['daemon']
+        daemon: options['daemon'],
+        cli_options: options.to_hash.symbolize_keys
       }
 
       puts '== The Middleman is loading'
--- a/middleman-core/lib/middleman-core/application.rb
+++ b/middleman-core/lib/middleman-core/application.rb
@@ -193,6 +193,7 @@ module Middleman
     define_setting :watcher_disable, false, 'If the Listen watcher should not run'
     define_setting :watcher_force_polling, false, 'If the Listen watcher should run in polling mode'
     define_setting :watcher_latency, nil, 'The Listen watcher latency'
+    define_setting :cli_options, nil, 'Options passed in from the cli.'
 
     # Delegate convenience methods off to their implementations
     def_delegator :"::Middleman::Logger", :singleton, :logger
--- a/middleman-core/lib/middleman-core/preview_server.rb
+++ b/middleman-core/lib/middleman-core/preview_server.rb
@@ -149,6 +149,7 @@ module Middleman
           config[:https]           = opts[:https] unless opts[:https].nil?
           config[:ssl_certificate] = opts[:ssl_certificate] if opts[:ssl_certificate]
           config[:ssl_private_key] = opts[:ssl_private_key] if opts[:ssl_private_key]
+          config[:cli_options]     = opts[:cli_options]
 
           ready do
             unless config[:watcher_disable]


balthisar:
Well, let me just close this until I figure out what went wrong!

balthisar:
Well, now that Travis-CI is happy, let me present a justification and use case for this proposed change.

*Justification*: Although _middleman_’s configuration is specified in `config.rb`, it’s common to specify additional configuration from the command line on an as needed basis. Although this change doesn't do that, it _does_ enable extension developers access to the Thor options specified on the command line anywhere app.config is available. Command line configuration simply becomes an extension of the `config.rb` configuration.

*Sample Use Case*: An extension developer wants to develop an extension that adds a command line argument. He can already override the Build and Server cli-classes to specify additional Thor options, but has no way to access these options within the scope of his extension (short of an external workaround involving environment variables). With this change the extension can perform additional self-configuration using the command line arguments in, e.g., `after_configuration`.

Note that I chose to expose these as a hash rather than an object instance in order to keep consistency between Build and Server; the latter uses a hash when starting the server.

tdreyno:
I wonder if this couldn't be an even bigger feature (I'd do the dev if it makes sense). What if any Middleman config option could be set on the CLI?

stevenosloan:
Build directory or asset host maybe? Those are the only ones I can think of that we end up setting with ENV with any frequency. I could definitely see the case for extensions having options they would like to expose though -- quickly glancing at the extensions directory, for eg. if [middleman-php](https://github.com/lord/middleman-php#configuration) wanted to add a `--php-show-debug` flag to the preview server.

tdreyno:
We've definitely set `data_dir` with a ENV for stuff at work.

balthisar:
I'm glad you're receptive to the idea. I'll admit that I was simply trying to avoid using external environment variables in my extension and this seemed like the least intrusive way of doing so. Still, I'm overriding cli `Build` class in a wanky way to make it happen. Because it's a Thor::Group I have to alias `build` in a `no_commands` block, override `build`, and then call the original `build` after handling cli options (and similar in class `Server`). This is perfectly acceptable to me. _However_...

If we wanted to formalize the ability of extensions to handle cli parameters, I could try to introduce a new callback in Build/Server, once `@app` has been setup. I'm not sure I would want to attempt this, though; I'm not nearly familiar enough with the internals to stick to your architecture and design.

Alternatively we could break the existing build task into three parts. Instead of `build`, break it into `build_setup`, `extension_setup`, and then finally `build`. In core `extension_setup` will do nothing, but extensions can override it to perform whatever actions they need. What I don't know, though, is whether the methods' order is preserved when overriding a class (if not, the extension would have to override all three methods and do nothing but call super in two of them). On the other hand I think preserving the order is the purpose of Thor::Groups, isn't it? I will experiment.

If it's possible to formalize the added step with one of the above, then it's probably possible to stick to your modern convention of "everything is an extension" and simply write an extension to accept cli parameters to override the configuration. I seem to think that app.config is frozen to extensions, though, so maybe it would have to be something in core. I'll experiment with that, too.

balthisar:
Actually, `after_build` does seem to be the right place to make changes based on CLI, and of course config isn't frozen (I'm already using it in my extensions). I think the only real missing piece is the ability to handle _testing_ of CLI arguments before Build/Server with no other callbacks necessary, and I'd be happy doing this by overriding an otherwise NOOP Thor step before build/server.

In Server, I've verified the cli options are still available every time the app is re-created after config.rb changes and `after_build` is called again. Suppose:
- Your extension changes `build_dir` per a cli argument.
- You change `build_dir` in `config.rb`
- Middleman restarts the server.
- `after_build` is called again.
- `Your extension changes `build_dir` per the original cli argument.

I'm okay with this, but I don't know if other users would expect live changes in `config.rb` to override the command line.

balthisar:
I've added a dummy task to the Build and Server groups, which are meant to be overridden by extensions so that they can test for parameters that are applicable to them.

I've also made a quick example extension [here](https://github.com/balthisar/middleman-cliconfig) that defines a new `--build_dir` (or `-o`) cli argmument for both Server and Build.

- The sample extension overrides Build and Server to add the Thor class options, and to "check" parameters before accepting them (in event of a failure exit 1 would stop the tasks).
- In `after_configuration` the CLI parameter is actually applied.
- There's also a sample project included in repository -- it's just the default template with suitably modified `config.rb` and `Gemfile`.
- Of course the sample extension requires the changes in this PR in order to function.

If one wanted to turn this into a full-fledged extension with the ability to override any of the config settings from the command line, then I think it would be trivial. I'd even be willing to take a stab at it myself if someone wants to help me get started with adding tests (I only know unit testing, and only in Objective-C and PHP). Or, take it and turn it into something nice.

tdreyno:
Hi @balthisar, I'm going to need to dig in and see how feasible extending all this is. I think it looks good, but I probably won't get a chance until the weekend.

balthisar:
@tdreyno, take your time. The current approach I've implemented is _bad_. I'll have to try to implement it as a callback, or even just see if `after_configuration` is good enough. Doing CLI checks by overriding a Thor step is a bad idea... multiple extensions will conflict if they try to preprocess their CLI parameters in the same overridden method.

balthisar:
I was overthinking it. Just the minute changes in this PR, and a really simple [sample extension](https://github.com/balthisar/middleman-cliconfig) using `after_config` and defining any new class_option overriding `Build` and `Server` should handle nearly 100% of use cases.

Multiple extensions adding the same class option or alias is undefined, unless Thor defines it.

tdreyno:
Okay, check out that change

balthisar:
@tdreyno, oh, that's _very_ nice. Every option is available from the cli now, and it's built-in rather than an extension.

But how do I define my own cli arguments now? If I override `build.rb` and add a new class option, it's not visible anywhere. In `build.rb:68` `config.setting()` is returning nil for my new cli argument. I've _also_ defined the cli argument in my extension as an option. 

`config.setting()` is from `configuration.rb`, which says in the header:
~~~ ruby
# A class that manages a collection of documented settings.
# Can be used by extensions as well as the main Middleman
# application. Extensions should probably finalize their instance
# after defining all the settings they want to expose.
~~~

So what's the right way now to ensure that my custom cli argument is a member of the ConfigurationManger::settings hash?

* * *

On a separate note the cli now creates a throwaway instance of an app, meaning that callbacks are executed twice. This is no problem as `:exit_before_ready` is set, but that's not actually in the user-facing documentation, so extensions not checking for this setting might do complicated things unnecessarily the first time the callback is executed.

balthisar:
I see what's going on now; Middleman is only taking into account its own cli arguments; they're simply discarded in `build.rb` if there's not a Middleman configuration option that matches, even when adding new options by extending the Thor::Group.

And because these checks are being done in the `::Middleman::Application.new` block, callbacks are already executed. This is kind of a catch-22: if `config.settings()` returns nil we can't poll the extensions because they don't really exist as activated until .new is complete.

While my PR spurred an _awesome_ new feature, the change as-is doesn't address the original intention of my PR, which was to allow extensions to look for and act on cli arguments.

I'm looking at `build`, but presumably `server` is similar. During app creation, could we simply make the cli options visible to extensions? My PR has the unfortunate effect of lumping in the cli options as a single Middleman option, but it was simple. Perhaps a new app property? I could draft something up.

tdreyno:
Right now, it isn't exposing extension options. But, extensions can expose global options (new in v4, not documented yet): https://github.com/middleman/middleman/blob/master/middleman-core/lib/middleman-core/renderers/sass.rb#L14

Extensions should be able to be initialized multiple times. This will also happen when `config.rb` is change and Middleman needs to reload.

Rewarding your point about the original PR, we should definitely expose the CLI options globally.

balthisar:
@tdreyno,
> Right now, it isn't exposing extension options. But, extensions can expose global options (new in v4, not documented yet)

This is awesome, too! And it works from an extension. However settings made specifically in `config.rb` override any cli arguments given.

In `application.rb:302` the call to `instance_eval` overwrites everything gathered in the CLI. This is called from `init`'s `evaluate_configuration!` _after_ the call to the block is executed.

So right now the CLI can override configuration _defaults_, but not configuration settings from `config.rb`. I would suggest that there's a valid case for correct this. Maybe _normally_ I want `build_dir` to be `path_to_qa_folder/` but sometimes, with a CLI override, I would want to set it to `path_to_production_folder/`.

* * *

On a side note, with the changes made to accommodate the CLI, it's no longer possible to conveniently override the Thor::Groups because extension classes are being required first. In the case of `build.rb` for example, my Middleman::Cli::Build is loaded _first_, meaning I've got to change my class declaration to `class Build < Thor::Group`, causing the native `Build` class to override _me_ instead of the other way around. This works well enough for adding CLI arguments, but now I can't override the `build` task because it doesn't exist yet at the time the extension is loaded.

For example, I used to be able to do this:

~~~ bash
no_commands do
   alias_method :method_orig, :build
end

def build
   # extra stuff here
   ...
   method_orig
end
~~~

If I try to be clever and require `middleman-cli/build` it fails because native `build` is depending on a module that defines a method in the actual console application!

While I'm certain that the ability to do what I describe was never considered a supported use case, it was quite a convenient means of being able to, e.g., build multiple times with a single call to Middleman.

* * *

tdreyno:
Okay, the discussion about whether config or CLI overrides is a good one. I think CLI wins, like you said. I'll look into adjusting that.

Eeek, that's some metaprogramming :) w.r.t. hacking `build`, is the goal to provide this modification in the gem for others to use? Could you not just create a new command, without hacking build. Sorry, if I lost the original goal in this thread. What are you doing to `build` that requires this wrapping?

balthisar:
> w.r.t. hacking build, is the goal to provide this modification in the gem for others to use?

Yup. Right now I get by with environment variables and my own CLI, whereas I really just want the users to do `middleman build --blah` instead of using my old script.

> Could you not just create a new command, without hacking build. 

I can probably do that; I kind of do that now. I'm trying to "fix" my hacky, stupid, written-before-I-knew-anything-about-Ruby extension so that it's a _Middleman_ extension, and not some standalone thing that just happens to use Middleman.

> Sorry, if I lost the original goal in this thread. What are you doing to build that requires this wrapping?

I'm trying to introduce multiple targets without involving environment variables, using the normal `middleman build --target=my_target`; I could write my own command, but for existing Middleman users trying to piggyback on existing `build` seems more natural. Actually for a single target, that's elementary and I don't need to override anything. However if I specify `build --target=all` then I want to build every target, hence overriding the Thor action.

My current options are
- Just add a `buildall` command, in line with what you suggest.
- Override `build` with my funky metaprogramming (broken currently).
- Simply add support for multiple targets in a fork and hope you'll accept a pull request (with the new CLI support, this looks to be trivially easy).

In any case none of this is a deal killer for me and I'm having a blast seeing what I can force Ruby to do.

In any case, the _original_ intention was to expose CLI arguments to app.config so I could see them in my extension's callbacks, and with your changes that works (except for the case they're being re-read from config.rb currently).

tdreyno:
Wonder if it's possible to inherit our build, then re-register yours with thor over the top?

balthisar:
> Wonder if it's possible to inherit our build, then re-register yours with thor over the top?

I'll continue to work with it and see what I can do. In the meantime I'll move the CLI priority issue over to #1843, and close this PR.

PR #1835, opened 2016-03-16
Avoid prefixing assets with asset dirs if no such file is on disk.

puyo:
I think the `path` local variable is shadowing the argument `path` and resulting in behaviour I'm not fond of: prefixing assets with asset dirs even when those assets don't exist. This breaks my middleman blog, in which I put HTML templates and images that all pertain to a single blog entry into a directory together.

### DIFF
--- a/middleman-core/lib/middleman-core/util/paths.rb
+++ b/middleman-core/lib/middleman-core/util/paths.rb
@@ -80,10 +80,11 @@ module Middleman
       result = if resource = app.sitemap.find_resource_by_destination_path(url_for(app, path, options))
         resource.url
       else
-        path = ::File.join(prefix, path)
-        if resource = app.sitemap.find_resource_by_path(path)
+        prefixed_path = ::File.join(prefix, path)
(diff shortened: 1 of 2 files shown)


tdreyno:
I'm not entirely sure it makes sense for the behavior of this method to change based on that context. Would something like `<%= image_tag "./relative-image.png" %>` work instead (or the relative path in the markdown image tag if thats what you're using?)

puyo:
Yeah you're right and I was going to come back and close the PR and go and fix it in a way that makes more sense.

My idea is that the result of the `url_for` method is being passed into `find_by_destination_path`, but in my case the result of `url_for` was relative to the current resource (e.g. '../articles/article1/image.jpg') but all the keys used to look up destination path are absolute paths (e.g. '/articles/article1/image.jpg') so the resource was not found even though it does exist on disk. I think anything passed into `find_by_destination_path` should be normalised into an absolute path. I'll check about and see if there's a change that is more logical.

tdreyno:
That seems like a better approach.

Issue #1833, opened 2016-03-14
Proxy *may* be dependent on sitemap order


stevenosloan:
I'm not sure if this is really an issue or not -- I can't seem to reproduce it w/ any kind of sane limited case (the instance I'm running into it we're doing quite a bit of sitemap manipulation). But, if you manage to get in a state where the `ProxyResource` is ahead of it's target -- the build will fail with the `proxies to unknown file` error.

Here's an example I *was* able to create by manually resorting proxy resources to be first:
- [Example](https://github.com/stevenosloan/tmp_middleman_proxy_test)
- [Failure log](https://travis-ci.org/stevenosloan/tmp_middleman_proxy_test/builds/115917341)

I can probably find a workaround for my case, but if you think this is an actual issue I'm happy to dive in on it -- would just like to know where/how you think it should be handled.

cheers,
steven

tdreyno:
I thought this should have fixed it: https://github.com/middleman/middleman/commit/0d983231d0ab90e64e6a96da06d9fe62c8f28fab

Maybe you just need to run your manipulator later?

tdreyno:
Basically, that stuff has 2 passes.

1. On config load
2. On invalidation of loops in config (if the value of data.whatever changes)

That second form runs near the end of the sitemap manip run. It's been causing issues.

stevenosloan:
I was able to fix it for now by changing the filename we were proxying to from `doc_page.html.haml` => `01-doc_page.html.haml`. I'll keep playing with it and see if I can get a reduced test case working (that isn't just an attempt to sabotage the order there)

Here's an output from the real case if it's any help: https://gist.github.com/stevenosloan/d85df22ac530ff03c1b8

stevenosloan:
Closed with #1836

PR #1836, opened 2016-03-16
Fix bug with store rebuilding and proxy resources.

puyo:
I was trying to make `middleman-blog` work with middleman version 4.1.2 today. This is the code change I had to make in order to make the two compatible.

The call to `resource.page_id` for a proxy resource involves call to the proxy target's `metadata` which involves looking up the resource by `@store.find_resource_by_path` which fails because `@_lookup_by_path` is not finished re-building at the time of the call.

Splitting these three steps out fixes this problem because `@_lookup_by_path` is complete by the time the proxy resource's `metadata` method is ever called.

I will come back and write a test if that's what it takes to get this PR included. :)

### DIFF
--- a/middleman-core/lib/middleman-core/sitemap/store.rb
+++ b/middleman-core/lib/middleman-core/sitemap/store.rb
@@ -237,7 +237,16 @@ module Middleman
                 # Rebuild cache
                 @resources.each do |resource|
                   @_lookup_by_path[resource.path] = resource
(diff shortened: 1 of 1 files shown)


stevenosloan:
@tdreyno FWIW this fixes #1833 for me

tdreyno:
Awesome. Thank you!

PR #1842, opened 2016-03-19
Fix asset_url and relative option bug.

puyo:
If the relative option is set and the asset is stored in the directory
alongside the resource referencing it, asset_url produces the
wrong path ('../images/image.jpg' rather than just 'image.jpg').


tdreyno:
Lovely.

Issue #1852, opened 2016-03-22
asset_hash rewriter doesn't rewrite specific CSS files with `content: " "` 


bradgessler:
Given the CSS file `./source/stylesheets/bug.css`:

```css
.no-bug{background-image:url(/images/middleman-logo.svg)}
.bug{content:"";background-image:url(/images/middleman-logo.svg)}
.no-bug{content:""; background-image:url(/images/middleman-logo.svg)}
```

and the `config.rb` file:

```ruby
page '/*.xml', layout: false
page '/*.json', layout: false
page '/*.txt', layout: false

# Build-specific configuration
configure :build do
  activate :asset_hash
end
```

When I request the resource `http://127.0.0.1:4567/stylesheets/bug-ff7fb4c7.css`:

```css
.no-bug{background-image:url(/images/middleman-logo-257e02f7.svg)}
.bug{content:"";background-image:url(/images/middleman-logo.svg)}
.no-bug{content:""; background-image:url(/images/middleman-logo-257e02f7.svg)}
```

The url in the `.bug` class doesn't get rewritten because of the `content:""` attribute without a following space.

This problem exhibited itself when an asset with a `content: ""` attribute in a class comes before any other attribute.

I suspect the regexp at https://github.com/middleman/middleman/blob/master/middleman-core/lib/middleman-core/util/rack.rb#L25 is the source of the issue. Working through that expression ATM to figure out what its trying to parse.

bradgessler:
I was working with @adam-h on this problem and he pointed out the issue in the regexp:

```ruby
([\'\"\(,]\s*|# sourceMappingURL=)([^\s\'\"\)>]+((?-mix:\.svg)))
```

should have `(` added to the "any characters except .." clause of the regexp. That will look something like

```ruby
([\'\"\(,]\s*|# sourceMappingURL=)([^\s\'\"\)\(>]+((?-mix:\.svg)))
```

Patch on its way by one of us.

bradgessler:
I also add that this issue crops up when `css_minification` and `asset_hash` is enabled in `config.rb` and a css file as follows is in a project:

```sass
```

adam-h:
I submitted a patch for this, my test case uses pre-minified CSS code to highlight this bug (and minimize the patch by reusing that test app)

PR #1853, opened 2016-03-22
Fix asset_hash with content rule and uncommented CSS url

adam-h:
Fixes middleman/middleman#1852


tdreyno:
This is great, thanks. That regex is a beast. I want to refactor it into format-specific regexes (HTML, CSS, JS, etc) at some point. For now, this is very helpful.

adam-h:
Yeah, took me a little while to figure out exactly what it was doing and where it was going wrong. That sounds like it will be quite a good refactor when it lands, things will be simpler in those parts.

PR #1855, opened 2016-03-22
Add exts to asset_host extension

bradgessler:
Copied the extensions from `asset_hash` helper.

Not a fan of how these extensions are copy and pasted between extensions. I'd like to move `sources` and `exts` into a constant in a different PR so these don't drift apart again. Where is the preferred place to put that constant?


tdreyno:
Can you rebase?

I'm fine if you store that constant inside `::Middleman::Application`

bradgessler:
Rebased.

I'll add "constants" via `define_setting` in a different PR.

PR #1862, opened 2016-03-25
Fix server CLI

balthisar:
While the CLI options are working nicely for build, they're currently broken for server, in that `app.config` doesn't reflect anything in `after_config`.

This fix is identical to what's done in `build`, and properly applies things in the server environment.


tdreyno:
Whoopsie

PR #1872, opened 2016-04-07
Fix parent returns nil if path name changed.

balthisar:
Given that parent uses path (source path) for looking for file,
:38 @store.find_resource_by_destination_path() fails if the destination
path has changed. Using find_resource_by_path() ensures that we return
a parent resource.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
+.byebug_history
 npm-debug.log
+manifest.yaml
 /.bundle
 .DS_Store
 coverage
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 master
 ===
 
+# 4.1.5-4.1.6
+
+* Fix file recursion when looking for possible asset dependencies. Major preview server performance improvement.
+
 # 4.1.4
 
 * Unify default extensions for all URL processing extensions. #1855
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 master
 ===
 
+# 4.1.7
+
+* Upgrade fastimage to 2.0
+* Fix shutdown of external_pipeline commands when config.rb is changed. #1877
+* Allow calls to `app.` to work as collections after initial config parse. #1876
+
+
 # 4.1.5-4.1.6
 
 * Fix file recursion when looking for possible asset dependencies. Major preview server performance improvement.
--- a/Gemfile
+++ b/Gemfile
@@ -5,9 +5,7 @@ gem 'rake', '~> 10.3', require: false
 gem 'yard', '~> 0.8', require: false
 
 # Test tools
-gem 'pry', '~> 0.10', group: :development, require: false
-gem 'pry-byebug'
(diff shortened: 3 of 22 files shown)


tdreyno:
Can you add a test case for the currently failing condition?

balthisar:
> Can you add a test case for the currently failing condition?

Yup. Good chance to finally learn how testing works in Ruby. But this is the damndest thing: the bug this PR fixes is actually dependent on _how_ I set the resource's `destination_path`. If I simply:

~~~ ruby
resource.destination_path = new_output_path
~~~

…then this bug is exhibited; `parent` returns `nil` meaning `siblings` returns `nil`.

On the other hand, if I:

~~~ ruby
resource.destination_path.sub!(/something/, 'something_else')
~~~

Then all is well!

balthisar:
@tdreyno, I've added #1874 based on Master with a failing test case. I've actually appended this to the existing `sitemap_traversal.feature` since it seems that the failure is simply an extension of that case.

It doesn't fail when run under this PR, of course.

balthisar:
Right now `parent` uses the source `path` internally for all of its checks, except in :38 where we `find_resource_by_destination_path` (which fails in the expanded test case introduced in #1874).

This PR as it is works in all of the existing test cases, plus the new one that's introduced.

(from #1874):
> Seems like we should be unifying on destination_path, rather than source

This makes sense logically, except it requires rewriting `parent` completely to handle `directory_indexes` being used. Essentially we'd have to check for `directory_indexes` and disassemble the `destination_path`, leaving us back with pretty much the current state: the `path`.

- The PR in its current state leaves `parent` as a black box - internal implementation is private anyway.
- Passes `more-sitemap_traversal.feature`
- Passes the special case added to `sitemap_traversal.feature`
- Continues to use the original, source `path`.

tdreyno:
Why does directory indexes introduce a problem? The renamed path will be present in `destination_path` with the appended `/index.html`.

Do we need additional logic to say "the parent of index.html is the next level up index.html"? I would think that would already be happening.

balthisar:
> Do we need additional logic to say "the parent of index.html is the next level up index.html"? I would think that would already be happening.

Nope, you were right. I'd missed a destination_path earlier; all of the magic logic is already in place. This passes `sitemap_traversal.feature` (including the additional checks I added) and `more-sitemap_traversal.feature` which uses the directory_indexes.

tdreyno:
Right, but this test is wrong:

```
    When I go to "/rename_paths/item_three/sibling_one.html"
    Then I should see "Parent: rename_paths/30_item_three/index.html"
```

`30_item_three/index.html` is not a sibling of `item_three/sibling_one.html`

tdreyno:
The goal of this feature is to replicate something like the Apache file/directory browser

balthisar:
Indeed. I'll fix that. I'm about to go on a holiday, but I'll leave the PR open. If you get tired of it sticking around, feel free to close it, and I'll deal with a fix when I get back.

tdreyno:
no rush. Thanks @balthisar

PR #1874, opened 2016-04-09
Failing test for PR #1872.

balthisar:
Modify the feature `sitemap_traversal.feature` to include three more scenarios:
- The possibility that a file has been renamed for output.
- The possibility that a file is within a directory that has been renamed for output.
- The possibility that a file has been renamed for output, and is within a directory that has been renamed for output.

Master branch as of this PR _fails_ these last two tests; PR #1872 will cause these tests to pass however.

### DIFF
--- a/middleman-core/fixtures/traversal-app/config.rb
+++ b/middleman-core/fixtures/traversal-app/config.rb
@@ -1,3 +1,30 @@
+class MyFeature < Middleman::Extension
+
+  def manipulate_resource_list(resources)
+
+    resources.each do |resource|
+
+      # Contrived example: we're removing integer prefixes from file names
+      # during output, e.g., rename_paths/30_item_three/10_sibling_one.html.md
+      # becomes rename_paths/item_three/sibling_one.html
+
+      if ( match = /(\d+?)_/.match(resource.path) )
+        resource.destination_path = resource.destination_path.gsub(/(\d+?)_/, '') # fails!
+        puts resource.destination_path
+        # resource.destination_path.gsub!(/(\d+?)_/, '') # works!
+      end
+
+    end
+
+  end
+
+end
+
+::Middleman::Extensions.register(:my_feature, MyFeature)
+
+activate :my_feature
+
+
 proxy "/sub/fake.html", "/proxied.html", ignore: true
 proxy "/sub/fake2.html", "/proxied.html", ignore: true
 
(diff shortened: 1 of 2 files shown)


middleman-core/fixtures/traversal-app/config.rb
```diff
@@ -1,3 +1,30 @@
...
+    resources.each do |resource|
+
+      # Contrived example: we're removing integer prefixes from file names
+      # during output, e.g., rename_paths/30_item_three/10_sibling_one.html.md
+      # becomes rename_paths/item_three/sibling_one.html
+
+      if ( match = /(\d+?)_/.match(resource.path) )
+        resource.destination_path = resource.destination_path.gsub(/(\d+?)_/, '') # fails!
+        puts resource.destination_path
+        # resource.destination_path.gsub!(/(\d+?)_/, '') # works!
```
tdreyno: Can you explain what is going on here?

tdreyno:
I think this answers my questions regarding #1872. I think the current behavior is to be expected, rather than this behavior. Traversal is simple pathing between the output file paths.

What would be the goal of saying "file in folder X is a parent of a file in a completely different folder Y"?

middleman-core/fixtures/traversal-app/config.rb
balthisar: That bit changes the output path of various items, e.g. `rename_paths/30_item_three/20_sibling_two.html` becomes `rename_paths/item_three/sibling_two.html` on output.

By "contrived example" I mean a real application would be a bit more sophisticated, but an example might be using page numbers or articles dates.

balthisar:
> Traversal is simple pathing between the **output** file paths.

(Emphasis mine.) In this case though, `parent` isn't traversing the output path only. In `traversal.rb:18` the `root` path being compared is `path`, which is the source path, but on :38 we're comparing it against the output path and so it fails. Because it fails, `siblings` fails to return siblings, too.

> What would be the goal of saying "file in folder X is a parent of a file in a completely different folder Y"?

Oh, not, that's not the goal at all. In the test case the source folder is `30_item_three`, and the index is the parent to the two siblings in the same **input** folder. On output though, this folder is simply `item_three`, and the index is still the parent to the two siblings in the same **output** folder.

tdreyno:
Seems like we should be unifying on destination_path, rather than source

balthisar:
> Seems like we should be unifying on destination_path, rather than source

Let me rework #1872 to do that instead. From memory I think the probably reason it's the way it is the code is simpler not worrying about the index file not being on the destination path. I'm not looking at code right now, just surmising that.

tdreyno:
I think this is closed by a recent PR: https://github.com/middleman/middleman/pull/2188

PR #1887, opened 2016-04-20
WIP - Rendering pipeline.

bradgessler:
I don't intend to merge this yet; I'm opening this PR as a proposal and as a way to ask questions as they come along.

The motivation for a rendering pipeline is because I don't think Rack is the best place to do `asset_hash`, `asset_host`, and other content substitution for a rendered page because:

1. The HTTP context is not needed. e.g. we can assume all requests to Middleman are `GET` requests.
2. Its difficult to access the `Resource` object from the Rack context. In the case of the inline_url_writer, which is used by the asset extensions, Middleman has to jump through a bunch of hoops to figure out if an asset is of a certain type. For example, the Rack middleware looks at file extensions to determine if something is a css asset. Wouldn't it be cleaner to just iterate through a list of resources and call `Resource#content_type`?

I also think this is a much cleaner API for extension developers to tap into so they don't have to understand all of the stuff going on in rack (don't forget to make the last element in the response array an array!)

Currently the code in this PR is capable of:

```ruby
# From config.rb ... a contrived example of URL rewriting
ready do
  sitemap.resources.each do |r|
    r.pipeline << (body) -> { body.gsub('https://non-cdn-host.com', 'https://cdn-host.com') }
  end
end
```

## Implementation plan

* [x] Get feedback from @tdreyno about the current approach and gain understanding of:
  * `Resource` life-cycle - Is `ready` the right block to configure the pipeline? Where should I add filters to the resource pipeline when the app boots? When a page reloads? When the config reloads?
  * `Pipeline` could be confused with `ExternalPipeline`. Bike shed for 2 seconds on naming (`Filters`?).
  * How can I make these filters performant? e.g. where can I memoize the pipeline? It gets wiped out every time a resource is reloaded.
* [ ] Write unit tests for the new `Pipeline` class.
* [ ] Pick a core extension to port over into the pipeline. The `InlineUrlRewriter` is pretty gnarly so I might start with that.
* [ ] If everything feels right, port over all middleman core extensions (and maybe a few contrib extensions, like `middleman-syntax`) to use the rendering pipeline.

### DIFF
--- a/middleman-core/lib/middleman-core/extensions/pipeline.rb
+++ b/middleman-core/lib/middleman-core/extensions/pipeline.rb
@@ -1,0 +1,19 @@
+class Middleman::Extensions::Pipeline < Middleman::Extension
+  expose_to_config :pipeline
+
+  def initialize(app, options_hash={}, &block)
+    super
+    @pipelines = []
+  end
+
+  def pipeline(&block)
+    @pipelines << block
+  end
+
+  def manipulate_resource_list(resources)
+    @pipelines.each do |pipeline|
+      pipeline.call resources
+    end
+    resources
+  end
+end
--- a/middleman-core/lib/middleman-core/pipeline.rb
+++ b/middleman-core/lib/middleman-core/pipeline.rb
@@ -1,0 +1,24 @@
+require 'forwardable'
+
+module Middleman
+  class Pipeline
+    extend Forwardable
+    attr_reader :filters
+    attr_reader :resource
+
+    def_delegators :filters, :<<, :push, :unshift, :insert, :shift, :pop, :first, :clear
+
+    def initialize(resource)
+      @resource = resource
+      @filters = []
+    end
+
+    def render(*args)
+      process resource.render *args
+    end
+
+    def process(body)
+      filters.inject(body){ |body, app| app.call(body) }
+    end
+  end
+end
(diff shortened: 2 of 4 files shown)


tdreyno:
Naming: Rewriter, Middleware, Filter, ContentFilter

I'm guessing a rewritten inline url rewriter could look something like:

HTML:

```
sitemap.resources
    .select { |r| extensions.include?(r.ext) }
    .each { |r| r.pipeline << make_rewriter() }
```

Where make_rewriter is the regex stuff. Would be nice to keep the functionality abstract so others could build filters which parse URLs from content.

As far as the response proc, how about we allow 2 options:

1. A proc, takes a string, returns a string.
2. A class instance which responds to `execute_filter` which takes a resource and a string, returns a string.

This will allow more complex filters, things that need shared state or caches.

Finally, when to do things. I would think it would still happen in a manipulation step. Inline in config.rb may have additional complications, we'll just have to see.

Thanks for diving into this!

bradgessler:
Thanks @tdreyno! This is going to be a slow roll since I just had a child 22 days ago. Feel free to close this PR if you want to keep this repo clean and I'll just re-open when I make some progress on this.

tdreyno:
Congrats! No rush, this can sit for a while. Maybe I'll contribute as well if I get some down time

tdreyno:
Exploring more, here https://github.com/middleman/middleman/commit/6b6bb2eaffca060f9d7643db13862e6d31ec96b3

tdreyno:
Closing in favor of: https://github.com/middleman/middleman/pull/1900

PR #1891, opened 2016-04-24
Incompatibility of listen with Ruby < 2.2

danielbayerlein:
The dependency "listen" in version >= 3.1 isn't compatible with Ruby < 2.2.

See https://github.com/guard/listen/pull/371

### DIFF
--- a/middleman-core/middleman-core.gemspec
+++ b/middleman-core/middleman-core.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
   s.add_dependency('memoist', ['~> 0.14'])
 
   # Watcher
-  s.add_dependency('listen', ['~> 3.0'])
+  s.add_dependency('listen', ['~> 3.0.0'])
 
   # Tests
   s.add_dependency("capybara", ["~> 2.5.0"])


danielbayerlein:
@tdreyno Now the build is 💚

tdreyno:
Did they really just remove support for an entire version of ruby in a minor version number bump? :(

e2:
@tdreyno -

> Did they really just remove support for an entire version of ruby in a minor version number bump? :(

An "entire" MINOR version bump of Ruby.

People mindlessly update gems (without upgrading Changelogs) but ironically they're very conservative when upgrading backward-compatible Ruby versions.

Long story short: according to semver it's fine. Impact: 0.

@danielbayerlein - wasn't compatible "de facto", so this is more of a bugfix (in dependency declaration).

Also, Ruby 2.2. and 2.3 are recommended. Also, Ruby 2.1 is potentially buggy with no support: 

"This means that after the release of 2.1.9 we will never backport any bug fixes to 2.1 except security fixes."

Please don't make me the enemy of the community all because clueless users are frustrated about upgrading to a recommended version of Ruby.

e2:
The Ruby community is broken and needs patching ;)

Please consider branching of for any (unreasonable IMHO) backports to Ruby 2.1 and switching to Ruby 2.2.

danielbayerlein:
Amen.

FYI: I'm using Ruby 2.3 but not all users of Middleman do this.

PR #1900, opened 2016-04-28
Replace Rack middleware with Resource Filter Pipeline

tdreyno:


tdreyno:
Example usage:

```
change_spelling = -> (body) { body.gsub("hello", "hola") }

resources.filter { |r| r.locale == :es }.each do |r|
  r.filters << change_spelling
end
```

dg-ratiodata:
@tdreyno

So "rack" support is totally gone then? I use `middleman` to build my `proxy.pac`-files (Proxy Auto Configuration) with `proxy_pac_rb`.

Will this work? https://github.com/fedux-org/proxy_pac_rb/blob/master/README.md#using-rack-middleware-with-middleman

If rack-support is removed, isn't that a breaking "api" change?

tdreyno:
@dg-ratiodata That's why this is open to discussion. Rack will be removed from effecting build output. It is still present for dev mode for a variety of debugging tools and simulating a real hosting environment (cache headers, redirects, etc).

A non-Rack version of the above would look like:

```
compressor = ProxyPacRb::ProxyPacCompressor.new
loader = ProxyPacRb::ProxyPacLoader.new

resources.filter { |r| r.ext === '.pac' }.each do |r|
  r.filters << -> (content) {
    proxy_pac = ProxyPacFile.new(source: content)
    loader.load(proxy_pac)
    compressor.compress(proxy_pac)
  }
end
```

dg-ratiodata:
@tdreyno Thanks a lot for the feedback and the provided solution for my use case. :smile:

tdreyno:
@dg-ratiodata I think it's a way to make a common use case (manipulate on specific page, or a set based on some criteria) simpler to execute. Meanwhile, the support of a server framework to accomplish that in a static system is confusing, unless you know all the limitations going in.

The Rack interface is weird, requires a knowledge of Rack's streaming system, HTTP codes and the specific way it chains requests.

Where as "find the files, add a callback" sees conceptually simpler.

I started this branch to see how the internal Rack-based code would look using filters and have really like it. We'll keep Rack for server mode, but try to be explicit about disabling it for build mode (which was always a half-truth anyways).

dg-ratiodata:
Thanks for explaining that.

tdreyno:
Again, not committed to this yet. Will have a long hard think

sandstrom:
Seems like a good improvement! 👍  (saw your request for comments on Twitter)

Ch4s3:
So what's up with this? I really want to sync my articles to Algolia, but I can't seem to make it work as Rack middleware, and this looks great!

tdreyno:
@Ch4s3 Basically, this touches so much code that has been around forever that I'm hesitant to do it on it's own. Was hoping some other big features would come in and I could group them all as version 5.

That said, I don't think this should be used for pushing content to Algolia. Honestly, Rack doesn't make much sense either. Rather, a simple `after_build` hook for a Rake task would be better IMO.

PR #1901, opened 2016-04-29
Use https:// to clone templates on init

dg-ratiodata:
## Summary

This PR replaces git://-protocol in the `middleman init/new`-command through https:// to make it work via HTTP proxies. 

## Details

* Using git://-protocol for cloning git repositories does not work via
HTTP proxies. By replacing it with https:// instead it works for more
users. 

* HTTP proxies are quite common in enterprise environments so this PR reduces the barrier for such users.

## Motivation and Context

I would like to use `middleman` behind an HTTP proxy which fails on `middleman init my_project` since it's no possible for git to clone the template.

## How Has This Been Tested?

I ran your test suite for cli_init.feature and it succeeded on my local setup `ruby 2.3.x`.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

### DIFF
--- a/middleman-cli/lib/middleman-cli/init.rb
+++ b/middleman-cli/lib/middleman-cli/init.rb
@@ -113,7 +113,7 @@ module Middleman::Cli
     end
 
     def repository_path(repo)
-      repo.include?('://') || repo.include?('git@') ? repo : "git://github.com/#{repo}.git"
+      repo.include?('://') || repo.include?('git@') ? repo : "https://github.com/#{repo}.git"
     end
 
     # Add to CLI


tdreyno:
I think this is reasonable. If users prefer the old format, then a full path will still work.

maxmeyer:
@tdreyno Thanks for that.

PR #1907, opened 2016-05-06
Fix contract for Sitemap::Store.register_resource_list_manipulator

PiotrMisiurek:
Contract said that optional param priority can be only a number.
But the code also handle the boolean as a value to be compatibile
with old versions

Change contract to accept both Num and Bool


tdreyno:
Thank you

PR #1911, opened 2016-05-10
fix method missing `try`

matthewlehner:
Just started up a new middleman app today and it looks like `try` isn't being included from the ActiveSupport core extensions, so when I turned on the minify JavaScript and CSS options the build would fail.

### DIFF
--- a/middleman-core/lib/middleman-core/extensions/minify_css.rb
+++ b/middleman-core/lib/middleman-core/extensions/minify_css.rb
@@ -1,3 +1,4 @@
+require 'active_support/core_ext/object/try'
 require 'memoist'
 require 'middleman-core/contracts'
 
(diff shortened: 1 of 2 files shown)


tdreyno:
Thank you. Didn't realize that was a back port :(

olivierlacan:
This is broken for me in 4.0 and 4.1. Build won't even run since https://github.com/middleman/middleman/blob/3e4187568bc363ab9d3fb37bea0afcbe09cf598e/middleman-core/lib/middleman-core/extensions/minify_javascript.rb#L57 tries to call `try` on the header string.

olivierlacan:
@tdreyno 2.3.1

tdreyno:
Error message?

olivierlacan:
Same as this issue's original post. Undefined method try on String instance.

tdreyno:
Odd. Can you post your `Gemfile.lock`? Tests on that version of ruby was working fine.

PR #1912, opened 2016-05-13
allow partial lookups without a current_resource

stevenosloan:
I can't think of anything valid reason other than my specific case, but requiring a current_resource to find a partial just to support relative lookups doesn't seem necessary so I don't think it will hurt to loosen this up a bit.

In my case I found the issue trying to [support rendering partials in a required file](https://github.com/stevenosloan/middleman-sprockets/blob/e8b5cdeab528a48410b40f864324df59ed9e2732/features/test_cases/partial_rendering.feature#L24-L47). As a contrived example, with a file structure something like this:

```
source/
+- javascripts/
|   +- main.js
|   +- lib/
|   |   +- _templates.js.erb
+- partials/
|   +- templates/
|   |   +- _header.mustache.erb
```

Where
- `main.js` has `//= require '_templates'`
- `_templates.js.erb` has `<%= partial 'partials/templates/header.mustache' %>` somewhere in it

In the sprockets context, when `_templates.js.erb` is being rendered it doesn't know that it's inside the context of `main.js`, so it's current_path doesn't match up with a resource in the sitemap -- but since that partial is relative to source could still be found w/o an issue.

### DIFF
--- a/middleman-core/lib/middleman-core/template_context.rb
+++ b/middleman-core/lib/middleman-core/template_context.rb
@@ -127,30 +127,36 @@ module Middleman
     # @return [String]
     Contract String, Maybe[Bool] => Maybe[IsA['Middleman::SourceFile']]
     def locate_partial(partial_path, try_static=true)
-      return unless resource = sitemap.find_resource_by_destination_path(current_path)
-
-      # Look for partials relative to the current path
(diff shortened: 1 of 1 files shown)


tdreyno:
Seems reasonable to consider the root a valid search path to include.

Issue #1920, opened 2016-05-23
Add support for Tilt 2


mfenner:
I am sure this has been discussed before, but I can't find the relevant Github issue. It would be great if middleman would support newer versions of Tilt than `1.4.1` by changing the relevant line in `middleman-core.gemspec` to `s.add_dependency('tilt', ['>= 1.4.1', '< 3.0'])`, the same as `padrono-helpers`. Tilt is currently at version `2.0.4`, and I am specifically interested in Pandoc support for markdown conversions that was added in `2.0.3`.

If it helps I can submit a pull request after fixing all tests.

tdreyno:
We _were_ locked by Padrino not supporting it yet... didn't realize they had upgraded.

I would love a PR and can work with you if you encounter any weirdness.

mfenner:
I think Tilt 2 support was added to Padrino in January: https://github.com/padrino/padrino-framework/issues/1789. Will try a PR, but I am sure I will need some help.

mfenner:
Created pull request #1921. Most tests pass, but some markdown configuration options are not passed correctly. 

@tdreyno please have a look, and let me know what else you need.

mfenner:
@tdreyno do you want me to fix the errors in the pull request #1921 first, or can you have a look now?

tdreyno:
Closing in favor of #1921, moving comments over there.

PR #1921, opened 2016-05-23
Enable Tilt 2 support. #1920

mfenner:
Relaxed requirements for Tilt to `['>= 1.4.1', '< 3.0']`. Check Tilt version and use `Tilt.default_mapping.lazy_map` instead of `Tilt.mappings` with Tilt 2. Drop `RedcarpetTemplate::Redcarpet2` which was removed in Tilt 2.

Some tests checking for markdown configuration options still fail.

### DIFF
--- a/middleman-core/lib/middleman-core/application.rb
+++ b/middleman-core/lib/middleman-core/application.rb
@@ -344,23 +344,11 @@ module Middleman
 
     # Clean up missing Tilt exts
     def prune_tilt_templates!
-      # If Tilt 1.4.x
-      if ::Tilt.respond_to?(:mappings)
-        ::Tilt.mappings.each_key do |key|
-          begin
-            ::Tilt[".#{key}"]
-          rescue LoadError, NameError
-            ::Tilt.mappings.delete(key)
-          end
-        end
-      else
-        # If Tilt 2.x
-        ::Tilt.default_mapping.lazy_map.each_key do |key|
-          begin
-            ::Tilt[".#{key}"]
-          rescue LoadError, NameError
-            ::Tilt.default_mapping.lazy_map.delete(key)
-          end
+      ::Tilt.default_mapping.lazy_map.each_key do |key|
+        begin
+          ::Tilt[".#{key}"]
+        rescue LoadError, NameError
+          ::Tilt.default_mapping.lazy_map.delete(key)
         end
       end
     end
--- a/middleman-core/lib/middleman-core/application.rb
+++ b/middleman-core/lib/middleman-core/application.rb
@@ -344,11 +344,23 @@ module Middleman
 
     # Clean up missing Tilt exts
     def prune_tilt_templates!
-      ::Tilt.mappings.each_key do |key|
-        begin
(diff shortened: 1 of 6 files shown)


tdreyno:
Is there anything to keep us from moving fully to v2 rather than backwards compat?

mfenner:
I was following what other Ruby libraries are doing:

* sinatra: `tilt < 3, >= 1.3`
* padrino-helpers: `tilt < 3, >= 1.4.1``

But not sure whether we really need support for tilt 1.x.

tdreyno:
Tilt 2 came out 2.5 years ago, I think we're good to upgrade. In fact, we may be one of the ones holding Padrino back from fully upgrading :)

mfenner:
OK, changed requirements to `~> 2.0`.

tdreyno:
Cool. Any issues getting those tests to pass (or removing them in the Redcarpet 1 case)?

tdreyno:
Ping @mfenner any movement on the remaining failing tests?

jennasalau:
Hey guys,

We noticed a major issue with our project going from mm 4.1.10 to 4.1.11.

Doing so updates us to Tilt 2 and Tilt 2 try's to handle our `.jsx` files... I don't want MM touching or even looking at those files as I have webpack on it! ?? Going back to 4.1.10 fixes our below issue.

How can i tell MM to ignore `.jsx` files on Tilt 2?

See our stack trace

```
bundler: failed to load command: middleman (/Users/jd/.gem/bin/middleman)
LoadError: cannot load such file -- babel/transpiler
  /Users/jd/.gem/gems/tilt-2.0.5/lib/tilt/babel.rb:2:in `require'
  /Users/jd/.gem/gems/tilt-2.0.5/lib/tilt/babel.rb:2:in `<top (required)>'
  /Users/jd/.gem/gems/tilt-2.0.5/lib/tilt/mapping.rb:243:in `require'
  /Users/jd/.gem/gems/tilt-2.0.5/lib/tilt/mapping.rb:243:in `block in lazy_load'
  /Users/jd/.gem/gems/tilt-2.0.5/lib/tilt/mapping.rb:241:in `each'
  /Users/jd/.gem/gems/tilt-2.0.5/lib/tilt/mapping.rb:241:in `lazy_load'
  /Users/jd/.gem/gems/tilt-2.0.5/lib/tilt/mapping.rb:216:in `lookup'
  /Users/jd/.gem/gems/tilt-2.0.5/lib/tilt/mapping.rb:154:in `[]'
  /Users/jd/.gem/gems/tilt-2.0.5/lib/tilt.rb:48:in `[]'
```

tdreyno:
Does `ignore "*.jsx"` work in `config.rb`?

Might make sense for us to add that by default, we already remove some of Tilt's other over aggressive filters... I think they do `.html` for some reason

theozaurus:
I've just run into the same issue. Adding `ignore '*.jsx'` does not make a difference. I'm using 4.2.1. Any ideas?

Ch4s3:
@theozaurus you can install `gem 'babel-transpiler'`

PR #1926, opened 2016-05-28
Prevent overwrite of Slim embedded options

### LINKED ISSUE #1152: Redcarpet & HAML not working correctly
Config

```
set :markdown_engine, :redcarpet
set :markdown, {smartypants: true, underline: true}
```

Haml

```
:markdown
  -- "foo" _bar_ *test*
```

In the generated HTML the markdown is processed but the custom options (underline & smartypants in this example) aren't applied.
Markdown files work fine, it just doesn't work within the `:markdown` HAML filter.

```
module Haml::Filters
  module Markdown
    include Base
    def render text
      md_options = {smartypants: true, underline: true}
      x = Redcarpet::Render::SmartyHTML.new md_options
      y = Redcarpet::Markdown.new x, md_options
      y.render text
    end
  end
end
```

This change I'm currently using adds the options, but isn't using middleman's custom RedcarpetTemplate which has many useful changes:
https://github.com/middleman/middleman/blob/v3-stable/middleman-core/lib/middleman-core/renderers/redcarpet.rb
and supports stuff like `middleman-syntax` within the markdown.

I think what's happening by default is something like this:

```
module Haml::Filters
  module Markdown
    include Base
    def render text
      ::Tilt['markdown'].new{ text }.render
    end
  end
end
```

But for some reason this isn't applying the custom options.

### LINKED ISSUE #1925: Slim::Embedded options being reset?
Using Middleman 4.1.8, I have this in my config.rb:

    require 'slim'
    require 'redcarpet'

    # ... 

    markdown_config = {
      :renderer            => CustomMarkdownRenderer,
      :fenced_code_blocks  => true,
      :autolink            => true
    }
    set :markdown_engine, :redcarpet
    set :markdown, markdown_config
    Slim::Embedded.set_options markdown: markdown_config

These custom options work fine with *.md files, but not with *.slim files.

When I inspect `Slim::Embedded.options` from a Ruby block in my Slim templates, I find that `Slim::Embedded.options[:markdown]` has been set to:

    :context=>#<#<Class:0x007f8728f42610>:0x007f872564d0f8 @app=#<Middleman::Application:0x70109242212760>, ...

Is Middleman reassigning `Slim::Embedded.options`? How can I prevent this so that Slim will use the Markdown options that I specify?

svicalifornia:
Fixes #1925, and may also fix #1152.

### DIFF
--- a/middleman-core/lib/middleman-core/renderers/slim.rb
+++ b/middleman-core/lib/middleman-core/renderers/slim.rb
@@ -12,13 +12,9 @@ class ::Slim::Template
 
   def initialize(file, line, opts, &block)
     if opts.key?(:context)
-      context_hack = {
-        context: opts[:context]
-      }
-
       ::Slim::Embedded::SassEngine.disable_option_validator!
       %w(sass scss markdown).each do |engine|
-        ::Slim::Embedded.options[engine.to_sym] = context_hack
+        ::Slim::Embedded.options[engine.to_sym][:context] = opts[:context]
       end
     end
 


svicalifornia:
Not quite ready for prime time: embedded Sass and SCSS not working correctly. Will fix and add tests.

Issue #1938, opened 2016-06-17
Frontmatter in partials displays in Middleman 4.1.8/4.1.9


jimothyGator:
If a partial contains YAML front matter, in Middleman 4.1.8 and 4.1.9 (but not 4.1.7 and earlier) will render that front matter verbatim.

Example:
### index.html.erb

```
Hello <%= partial 'world' %>
```
### _world.html.erb

```

---
foo: bar

---
world
```
### Expected Output

```
Hello world
```
### Actual Output

```
Hello ---
for: bar

---
world
```

jimothyGator:
Added failing test case: #1939

tdreyno:
Thanks for reporting and providing the case, but this is by design. Do not include front matter in partials.

jimothyGator:
That's unfortunate to hear. We use Middleman to prototype sites which will be built on another platform. We use front matter to include meta data, from which we generate stubs for the production components. Prior to MM 4.1.8, this process worked.

tdreyno:
Can you show me what this looks like in practice?

jimothyGator:
It's a somewhat complex setup, and one that is particular to our use, but I'll do my best to explain it.

The final websites run in a CMS (Adobe Experience Manager [AEM], which is Java based), but before we build the components there, we build prototypes using Middleman, because this allows us to iterate much faster (particular because there is no compile and deploy cycle, as there is with AEM. This allows us to quickly create, debug, and refine our JavaScript and CSS/Sass; the same JavaScript and CSS files are shared between the prototype and the production site.

AEM components involve a number of files, much of it boilerplate. We've hooked in some Grunt tasks that generate these boilerplate files from the front matter we include in our Middleman partials. For each AEM component, we create a corresponding partial.

For example, we might define a header component in a partial named `_header.erb`, with the following front matter:

```
---
componentTitle: header
componentDescription: Our Site Header
componentGroup: Our Site
---
```

This would generate a variety of files for AEM: `Header.java`, `header.html`, `dialog.xml`, '.content.xml`, and the correct directory/package structure for them. The front matter values are used to create these files, which serve as starting points for the components. Developers then complete and customize the files. The code generation just provides a starting point, and automates a bit of the boilerplate and tedium away.

This is hardly a universal use case, to be sure, but others could probably find other uses for front matter, which was possible prior to 4.1.8.

tdreyno:
Makes sense. Why does this component need to be a partial?

jimothyGator:
[...]

PR #1939, opened 2016-06-17
add test case for frontmatter in partial

jimothyGator:


jimothyGator:
Failing test case for #1938

PR #1949, opened 2016-07-01
Add prefix option to asset_hash

adam-h:
This allows manually changing the filename so that fiel header changes
can be reflected on the CDN. E.g. if you turn on crossOrigin serving
(CORS) the asset hash doesn't change, but the CDN cache needs to be
broken in order to pickup the new header.


adam-h:
Thanks! Any chance of a release soon?

PR #1967, opened 2016-08-02
Updating to Active Support version 5

pas256:
As Rails version 5 is out of beta now, we can start using an updated version of Active Support.

I ran `bundle exec rake test` before and after, and nothing changed so this was surprisingly easy.

### DIFF
--- a/middleman-core/middleman-core.gemspec
+++ b/middleman-core/middleman-core.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   s.add_dependency('dotenv')
 
   # Helpers
-  s.add_dependency('activesupport', ['~> 4.2'])
+  s.add_dependency('activesupport', ['~> 5.0'])
   s.add_dependency('padrino-helpers', ['~> 0.13.0'])
   s.add_dependency("addressable", ["~> 2.3"])
   s.add_dependency('memoist', ['~> 0.14'])


pas256:
This will also allow others to use middleman within a Rails 5 app, for API documentation, for example. It resolves this issue:

```
Bundler could not find compatible versions for gem "activesupport":
  In Gemfile:
    middleman (~> 4.0.0) was resolved to 4.0.0, which depends on
      middleman-core (= 4.0.0) was resolved to 4.0.0, which depends on
        activesupport (~> 4.2)

    rails (~> 5.0.0) was resolved to 5.0.0, which depends on
      railties (= 5.0.0) was resolved to 5.0.0, which depends on
        activesupport (= 5.0.0)
```

pas256:
The tests for Ruby 2.2+ pass. Not sure if support for Ruby < 2.2 is as important as it used to be.

tdreyno:
I like this. Need to do a little twitter survey to check Ruby versions

tdreyno:
Sweet

kjg:
Is there any reason to remove support for 4.2? can we just widen the accepted range instead of making it only 5+?

pas256:
@kjg That might work too. I never remember the pattern.

kjg:
considering Rails sometimes has breaking changes in "minor" version releases I'd suggest

```ruby
s.add_dependency('activesupport', ['>= 4.2', '< 5.1'])
```

kjg:
@pas256 would you like me to make a new PR, or do you plan on it?

pas256:
@kjg I wasn't planning on it as I have moved on to other things at this point.

I am not against it at all so go for it. It makes sense to me.

PR #1970, opened 2016-08-11
Eliminate quadratic behavior in SourceWatcher::poll_once!

bdash:
Both `Array#reject` and `Array#include?` have linear time complexity. This results in quadratic time complexity when `Array#include?` is called within `Array#reject`'s block. Using `Array`'s difference operator gives the same result in linear time.

In a site with ~14,000 files, this drops the time taken by some calls to `SourceWatcher::poll_once!` from ~15 seconds each to just a few milliseconds.


tdreyno:
You're amazing! Thank you!

PR #1971, opened 2016-08-11
Improve performance of ignoring files

bdash:
On a site with around 14,000 files and a handful of simple ignore patterns, I was seeing `IgnoreDescriptor::ignored?` account for approximately 6 seconds of middleman's initialization time. This change reduces that time to around 0.6 seconds by:

1. Determining the type of ignore pattern only once per pattern, rather than every time the pattern is used.
2. Caching the normalized paths on `Resource` and `SourceFile`.


tdreyno:
Looks like a possible issue on Ruby 2.2 test. 

Thanks so much for this

bdash:
The failure in Ruby 2.2 appears to be an existing, sporadic issue. I was able to hit it locally on master with Ruby 2.2.5p319, but have not seen it when testing my branch.

tdreyno:
Cool! I was mobile and didn't check the full error log.

PR #1974, opened 2016-08-19
Tilt 2

kjg:
- Require any version of tilt between 2 and 3
- Drop RedcarpetTemplate::Redcarpet2 which was removed in Tilt 2
-  Use `Tilt.default_mapping.lazy_map` instead of `Tilt.mappings`
- Make use of `Tilt.default_mapping.extensions_for`

Fixes #1921

### DIFF
--- a/middleman-core/lib/middleman-core/application.rb
+++ b/middleman-core/lib/middleman-core/application.rb
@@ -344,11 +344,11 @@ module Middleman
 
     # Clean up missing Tilt exts
     def prune_tilt_templates!
-      ::Tilt.mappings.each_key do |key|
+      ::Tilt.default_mapping.lazy_map.each_key do |key|
         begin
           ::Tilt[".#{key}"]
         rescue LoadError, NameError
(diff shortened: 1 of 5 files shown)


kjg:
I get the same error locally too on master: `Error: Could not locate partial: ./_state` when running the i18-test-app tests. The other most recent master run on travis seemed to pass fine though. I'm not sure how to proceed. Any thoughts @tdreyno?

kjg:
Ah, its the auto update of padrino-helpers to 0.13.3 that is causing these failures

kjg:
#1975 should fix the remaining failures here.

kjg:
@tdreyno This is now green, can you please consider merging it? Thanks!

kjg:
@tdreyno This is now green, can you please consider merging it? Thanks!

mfenner:
Thanks @kjg!

tdreyno:
Thanks @kjg, sorry for the late reply, been traveling.

Issue #1963, opened 2016-07-21
config[:partials_dir] is ignored?


equinusocio:
Inside my `config.rb` i put `config[:partials_dir] = '/app/partials'` (i tried also `partials/`), the partials folder is right inside the app (ex source) and it contains a `_meta.erb` file.
Now in my `app/layouts/layout.erb` i have `<%= partial "meta" %>`. When i run the middleman server i get this error `Error: Could not locate partial: /_meta` because it seems to search the file inside the root folder. If i use `<%= partial "partials/meta" %>` (without the config[:partials_dir]) it works.

I missed something? Thanks.

tdreyno:
It's removed from 4.x: https://middlemanapp.com/basics/upgrade-v4/

tdreyno:
Your solution (always including the full path) is recommends (and easier to follow without the magic IMO too).

equinusocio:
Ok thanks. So i can concatenate a config option inside the partial helper?

Something like 'partial config[:partial] + 'meta''?

vvasabi:
@equinusocio This is a hack (placed in `config.rb`) I use to bring back the v3 partial behaviour:

```ruby
helpers do
  def partial(template, options={}, &block)
    if Dir.exists? File.join(config[:source], config[:partials_dir], File.dirname(template))
      template = File.join(config[:partials_dir], template)
    end
    super(template, options, &block)
  end
end
```

Hope that helps.

equinusocio:
@vvasabi thanks. @tdreyno There is a specific reason of why this conf was removed? 

However, after trying several static sites generators, I'm delighted to middleman and I consider it the best and most complete one.

tdreyno:
@equinusocio We took a look during the planning for v4 at features which required either a lot of code or a lot of documentation. `partials_dir` was one that was surprisingly difficult to implement, since we don't actually own the template layer (we use the Padrino project). We decided the simplified code-base was worth being a little more explicit in the templates. And, after all, the old behavior is only a couple of lines away if you can code a helper for your specific functionality (rather than all possible user functionalities).

Sorry for the inconvenience, hope everything works well for you.

Issue #1732, opened 2016-01-05
partials_dir on config.rb not working


kisin:
my config has partials_dir set to "partials" folder
but when calling the partial method i must add the partials folder for it to find the file

  
config.rb
```
set :partials_dir, 'partials'
```

layout.erb (if i remove the partials folder it will not work)
```
<%= partial "partials/header-top" %>
```

joshukraine:
The `partials_dir` setting has been removed as of 4.0.0.beta.1:

> Remove the partials_dir setting. Partials should live next to content, or be addressed with absolute paths.

See [CHANGELOG.md](https://github.com/middleman/middleman/blob/master/CHANGELOG.md#400beta1).

kisin:
ok, good to know.

rtack:
this should be documented in the https://middlemanapp.com/basics/upgrade-v4/ page. Just spent half a day trying to figure out why it wasn't working anymore.

tdreyno:
Sure thing: https://github.com/middleman/middleman-guides/commit/324b035af07e57e43e8b2d7e2818799f71bad708

PR #1975, opened 2016-08-19
Move partials into partial dir for padrino 0.13.3 compatibility

kjg:


tdreyno:
Wait, so they changed something about partials in a patch release?

kjg:
I didn't spend too much time trying to figure out exactly which commit
changed the behavior. But moving this partial into the partial directory
seems to make specs pass again.

kjg:
Are you open to merging this?

tdreyno:
I'd like to avoid forcing people to move a bunch of their files. I'd really like to find another solution. But I haven't had enough time to check the latest Padrino changes.

kjg:
Do people not keep their partials in the partials dir? Either way, this PR
doesn't change that. This PR just fixes the current failing specs on the
**master** branch.

The other way we can fix the specs is by pinning padrino to < 0.13.3

Just let me know which way to go, but currently master has failing tests
and we should fix them.

kjg:
I believe this is the change https://github.com/padrino/padrino-framework/pull/2047/files but it ultimately shouldn't matter, **master** does not pass CI right now.

How would you like it fixed, we can pin padrino to < 0.13.3 or we can merge this PR or do something different

tdreyno:
Hmm, so maybe it's a load order thing? We're also overloading their render method to implement partial stuff.

Best case fix would be to support latest Padrino, but also let the current tests pass.

Sorry, but I'm completely slammed at work, so PR reviews are quite slow right now. Thanks for contributing.

kjg:
So it's expected that partials can just be put anywhere for middleman?

stevenosloan:
@kjg that's my understanding at least -- [a little](https://github.com/middleman/middleman/issues/1732) [backstory](https://github.com/middleman/middleman/issues/1963), the partials directory got dropped in v4.

Digging in on that Padrino change, it's pretty subtle -- but if a template *doesn't* have a directory in it's name there will be a '.' prefixed once the template_path gets joined up, which I *think* is throwing off the MM render method.

```
==== testing with path/thing.erb
-> output for old
   path: path
   name: thing.erb
   template_path: path/_thing.erb
-> output for new
   path: path
   name: thing.erb
   template_path: path/_thing.erb
==== testing with thing.erb
-> output for old
   path:
   name: thing.erb
   template_path: /_thing.erb
-> output for new
   path: .
   name: thing.erb
   template_path: ./_thing.erb
```

So in that specific spec, moving the called partials *next* to the files that were asking for them let the relative path lookup in locate_partial (in middleman) find them. I've gotta hop off to some other stuff, but I *think* changing the regex here https://github.com/middleman/middleman/blob/master/middleman-core/lib/middleman-core/template_context.rb#L132 to `/^\.?\//` so that it would also strip a leading '.' should fix things and be compatible w/ older Pandrino?

kjg:
@stevenosloan Yeah there is a couple of different places this could be accounted for in middleman, I was looking at making this change:

```diff
diff --git a/middleman-core/lib/middleman-core/template_context.rb b/middleman-core/lib/middleman-core/template_context.rb
index 5efd8c4..fa451a8 100644
--- a/middleman-core/lib/middleman-core/template_context.rb
+++ b/middleman-core/lib/middleman-core/template_context.rb
@@ -101,7 +101,7 @@ module Middleman
     # @return [String]
     Contract Any, Or[Symbol, String], Hash => String, Maybe[Proc] => String
     def render(_, name, options={}, &block)
-      name = name.to_s
+      name = name.to_s.sub(/^\.\//, '')

       partial_file = locate_partial(name, false) || locate_partial(name, true)
```

but maybe the right thing to do is to fix Padrino instead since this is a breaking change with them

stevenosloan:
Yeah fixing in Padrino seems like the best option to me, doesn't seem like that was an intentional change -- especially w/ one of the use cases being middleman (https://github.com/padrino/padrino-framework/issues/1872) ... and looks like you're already on it 👍

tdreyno:
Thanks for the digging @stevenosloan

kjg:
This has been fixed in padrino 13.3.3.1

tdreyno:
Awesome. Thanks!

PR #1976, opened 2016-08-19
Allow for activesupport 4.2 again

kjg:
There is no current reason to remove support for 4.2
This now makes for a backwards compatible change whereas before this would
have called for a new major version of middleman-core

Refs #1967


kjg:
Depends on #1975 for CI to pass

kjg:
@tdreyno Can you please consider merging this? Thanks!

PR #1984, opened 2016-09-10
Updates to support Rails 5, Rack 3 and Sinatra 2 (beta)

### LINKED ISSUE #1983: Rails 5 Compatibility Error
So I'm assuming this is a rare (or at least largely unsupported) use case for middleman from my inability to find it documented, but I have a very simple Rails 5 app (I am using RoR entirely because I am most comfortable with it and I wanted to take advantage of sprockets and such) and I'd like to get my feet wet with middleman to automate the creation of a static version of that app. (If RoR and middleman are completely incompatible and this is a fool's errand, please let me know, but I figured there would be some way to do it since RoR sits on top of `rack`, which *is* compatible.)

At first I ran into the issue that `middleman-core` had a dependency of `activesupport ~> 4.2`, but -- upon seeing that this dependency was changed in cbad571 (#1967), I attempted to redirect my `Gemfile` to the master branch of this GH repo, assuming that would solve the issue. 

Unfortunately, however, I quickly hit another roadblock in that [`actionpack ~> 5.0.0.1`](https://github.com/rails/rails/blob/v5.0.0.1/actionpack/actionpack.gemspec#L24) has a dependency of `rack ~> 2.0` and the HEAD [`middleman-core`](https://github.com/middleman/middleman/blob/197093b36cf3e93ec0fc35f18e33dc6bff75e87c/middleman-core/middleman-core.gemspec#L23) specifically requires `rack >= 1.4.5, < 2.0`. 

Is there any way to either remove the upper limit on the rack version requirement, or else make it `< 3` to be in line with `rails`'s assumption that it won't change in a backward-incompatible way until at least the next major version increment? 

Thanks,
Mike

mvastola:
- Bump upper boundary of version for Rack dependency from 2.0 to 3 (exclusive). 
  - This allows compatibility with `actionpack 5.0.x` (and thus `rails 5.0.x`) which requires `rack 2.x`, while preserving compatibility with `rails < 5`, which requires `rack 1.x`.
- Bump the minimum version of `sinatra` used in CI testing from `1.4` to `2.0.0.beta2` (inclusive).
See #1983.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,7 @@ gem 'kramdown', '~> 1.2', require: false
 gem 'slim', '>= 2.0', require: false
 gem 'liquid', '>= 2.6', require: false
 gem 'stylus', '>= 1.0', require: false
-gem 'sinatra', '>= 2.0.0.beta2', require: false
+gem 'sinatra', '>= 1.4', require: false
 gem 'redcarpet', '>= 3.1', require: false
 
 # Dns server to test preview server
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,7 @@ gem 'kramdown', '~> 1.2', require: false
 gem 'slim', '>= 2.0', require: false
 gem 'liquid', '>= 2.6', require: false
 gem 'stylus', '>= 1.0', require: false
-gem 'sinatra', '>= 1.4', require: false
+gem 'sinatra', '>= 2.0.0.beta2', require: false
 gem 'redcarpet', '>= 3.1', require: false
 
 # Dns server to test preview server
--- a/middleman-core/features/asset_host.feature
+++ b/middleman-core/features/asset_host.feature
@@ -25,7 +25,8 @@ Feature: Alternate between multiple asset hosts
     And a file named "config.rb" with:
       """
       activate :asset_host, host: Proc.new { |asset|
-        "http://assets%d.example.com" % (asset.hash % 4)
+        hash = Digest::MD5.digest(asset).bytes.map!(&:ord).reduce(&:+)
+        "http://assets%d.example.com" % (hash % 4)
       }
       """
     And the Server is running
--- a/middleman-core/lib/middleman-core/extensions/minify_css.rb
+++ b/middleman-core/lib/middleman-core/extensions/minify_css.rb
@@ -72,7 +72,7 @@ class Middleman::Extensions::MinifyCss < ::Middleman::Extension
       end
 
       if minified
-        headers['Content-Length'] = ::Rack::Utils.bytesize(minified).to_s
+        headers['Content-Length'] = minified.bytesize.to_s
         response = [minified]
       end
 
--- a/middleman-core/lib/middleman-core/extensions/minify_javascript.rb
+++ b/middleman-core/lib/middleman-core/extensions/minify_javascript.rb
@@ -64,7 +64,7 @@ class Middleman::Extensions::MinifyJavascript < ::Middleman::Extension
       end
 
       if minified
-        headers['Content-Length'] = ::Rack::Utils.bytesize(minified).to_s
+        headers['Content-Length'] = minified.bytesize.to_s
         response = [minified]
       end
 
--- a/middleman-core/lib/middleman-core/rack.rb
+++ b/middleman-core/lib/middleman-core/rack.rb
@@ -133,9 +133,15 @@ module Middleman
 
     # Immediately send static file
     def send_file(resource, env)
-      file      = ::Rack::File.new nil
-      file.path = resource.file_descriptor[:full_path]
-      response = file.serving(env)
+      file     = ::Rack::File.new nil
+      path     = resource.file_descriptor[:full_path]
+      if !file.respond_to?(:path=)
+        request  = ::Rack::Request.new(env)
+        response = file.serving(request, path)
+      else
+        file.path = path
+        response = file.serving(env)
+      end
       status = response[0]
       response[1]['Content-Encoding'] = 'gzip' if %w(.svgz .gz).include?(resource.ext)
       # Do not set Content-Type if status is 1xx, 204, 205 or 304, otherwise
--- a/middleman-core/middleman-core.gemspec
+++ b/middleman-core/middleman-core.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
 
   # Core
   s.add_dependency('bundler', ['~> 1.1'])
-  s.add_dependency('rack', ['>= 1.4.5', '< 2.0'])
+  s.add_dependency('rack', ['>= 1.4.5', '< 3'])
   s.add_dependency('tilt', ['~> 2.0'])
   s.add_dependency('erubis')
   s.add_dependency('fast_blank')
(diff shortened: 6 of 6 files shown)


mvastola:
@tdreyno So something really odd is going on here. The CI check for Ruby 2.2.4 is passing, but the checks for 2.3.1 and ruby-head are failing, yet my change is almost definitely not the cause of this issue.  

I cloned my fork onto my local machine, and upon running `bundle update`, I'm noticing that `gem 'sinatra', '>= 1.4'`, which is listed in the `Gemfile` has a dependency of `rack ~> 1.5`. So even though the `rack` dependency is satisfied as long as the version is `>= 1.4.5` and `< 3`, the additional constraint imposed by `sinatra` here is ensuring that the CI tests run with a `rack` version of `1.x` (in this case, `1.6.4`). 

Weirder still is *which* tests are failing. (In particular, note that the issues on `ruby 2.3.1` on Travis and `ruby 2.3.0` on my box have *no* crossover.) Here they are broken down by Ruby version:
* `ruby 2.3.1` (Travis)
  * `cucumber features/asset_host.feature:23` # Scenario: Set proc host with inline-option
* `ruby 2.3.0` (my machine)
  * `cucumber features/console.feature:3` # Scenario: Enter and exit the console
  * `cucumber features/relative_assets_helpers_only.feature:3` # Scenario: Rendering css with the feature enabled
  * `cucumber features/relative_assets_helpers_only.feature:91` # Scenario: URLs are not rewritten for rewrite ignored paths
* `ruby-head [2.4.0dev]` (Travis)
  * (There seems to be a pervasive error with `require`ing `coffee_script` that may be the source of these issues.)
  * `rspec ./spec/middleman-core/util_spec.rb:165` # `Middleman::Util::asset_url` returns a resource url when `asset_hash` is on
  * `rspec ./spec/middleman-core/util_spec.rb:157` # `Middleman::Util::asset_url` returns a resource url if given a resources path
  * `rspec ./spec/middleman-core/util_spec.rb:149` # `Middleman::Util::asset_url` returns a resource url if given a resource's destination path
  * `rspec ./spec/middleman-core/util_spec.rb:101` # `Middleman::Util::asset_url` returns path relative to the provided `current_resource`
  * `rspec ./spec/middleman-core/util_spec.rb:137` # `Middleman::Util::asset_url` when `relative` is `true` raises error if not given a `current_resource`
  * `rspec ./spec/middleman-core/util_spec.rb:117` # `Middleman::Util::asset_url` when `relative` is `true` returns path relative to the provided `current_resource`
  * `rspec ./spec/middleman-core/util_spec.rb:130` # `Middleman::Util::asset_url` when `relative` is `true` when the asset is stored in the same directory as `current_resource` returns path relative to the provided `current_resource`
  * `rspec ./spec/middleman-core/util_spec.rb:92` # `Middleman::Util::asset_url` when `http_prefix` is activated returns path with `http_prefix` pre-pended if resource is found
  * `rspec ./spec/middleman-core/util_spec.rb:96` # `Middleman::Util::asset_url` when `http_prefix` is activated returns path with `http_prefix` pre-pended if resource is not found
  * `rspec ./spec/middleman-core/util_spec.rb:196` # `Middleman::Util::find_related_files` Finds partials possible related to Scss files
  * `rspec ./spec/middleman-core/util_spec.rb:188` # `Middleman::Util::find_related_files` Finds partials possibly related to ERb files

Any advice/assistance/thoughts/theories/hypotheses are welcome. I'm a pretty seasoned ruby and rails dev, but totally new to middleman, so I'm probably not the best person to spot what might be going wrong here..

tdreyno:
Hmm, so Sinatra is still blocking us... 

Maybe we can start relying on their beta? https://rubygems.org/gems/sinatra/versions/2.0.0.beta2

mvastola:
Sure, but if the build is failing as it is, that suggests a pre-existing problem in the code base, right? (Worse still, there seems to be little rhyme or reason as to which CI tests are failing.) 

It'll be very difficult to see if changing the `sinatra` version will break anything (or if the `rack` version change did) until the pre-existing bugs are worked out, don't you think?

tdreyno:
Master build looks good to me: https://travis-ci.org/middleman/middleman/branches

mvastola:
True. Super odd, too. If I expand the `bundle update` in both mine and the master builds, it shows the exact same packages (and versions) so there's no accounting for this failure then. I'm going to force a rebuild here just in case before I go further, just out of curiosity.

tdreyno:
I can't seem to find anywhere the changelog for Rack 2... no idea what changing it would mean...

mvastola:
My bad. Here you go. 
https://gist.github.com/mvastola/a80211b6134afb34ba5bcd330ec1d98e

tdreyno:
Yeah, I saw that... it's fairly confusing. Especially since it never mentions a 2.0 release. It also doesn't mention any removals, only deprecations... so it should just work, but in that case, why the large version bump?

mvastola:
Re `rack 2`: Adding features? My best guess..
Re my change: Reverting it did fix things, but I have a lead on what happened. I think the error in the CI test on Travis we were seeing is a bug(ish) problem with the test and not a real issue at all. Will keep you posted.

mvastola:
Ok, so I have a feeling this will work. Tracking down the CI test that was failing after I changed the rack version (which actually caused no change), a rather un-robust implementation of setting up round-robin asset host servers was the culprit. 

The [previous code](https://github.com/middleman/middleman/pull/1984/commits/f8bd1808f728a099cffedd15ef8db6732059c96f#diff-8d310b8c4af8f81fde7852d6cf989524
) fell prey to an edge condition wherein, after my change, apparently the relevant asset path strings got stored *just* the right memory addresses that calling `#hash` on the variables in question never produced an integer which, when divided by 4, produced a remainder of 1. This caused the test to fail. I [fixed things](https://github.com/middleman/middleman/pull/1984/commits/f8bd1808f728a099cffedd15ef8db6732059c96f#diff-8d310b8c4af8f81fde7852d6cf989524
) by using a much more foolproof (although slightly clunky) digest method.

mvastola:
Anywho, this is ready to merge, unless you think the `sinatra` version bump should be part of this same PR. (It's also up to you if you want to officially support compatibility of your gem with a beta release.)

I don't know much about `sinatra` though, so I really wouldn't be able to debug any problems corresponding to that side of things. At the same time, without upgrading `sinatra`, we can't run the CI tests against `rake 2.x` (though according to the docs, nothing should be changed).

mvastola:
Actually, decided to push it in a separate commit (15fa65e) and we'll see if the CI tests pass and go from there.

mvastola:
@tdreyno I think this is all set (with `sinatra`). Check it out!

tdreyno:
If you revert the gem changes, but leave the code changes. Does everything pass? (trying to catch regressions, we should not break old rack support)

mvastola:
@tdreyno Ok. I think this is (finally, actually) good to go. It passes Travis with my code changes with both the old *and* new versions of `rack`/`sinatra`. When you merge you'll obviously want to only take the first commit (d3c3043) as the third one reverts the second.

Let me know if there's anything else, lest I missed something!

tdreyno:
Nice work, thank you!

PR #1996, opened 2016-10-19
Make .html implicit for page IDs

joallard:
Got surprised by this while working on #1466.

We also need to remove `localizable/` and the `.en` from the page id, but I don't know at what level we should do that, resource level or i18n, since this is kind of a shared thing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,11 @@ gem 'rspec', '~> 3.0', require: false
 gem 'cucumber', '~> 2.0', require: false
 gem 'addressable', '~> 2.4.0', require: false
 
+# Pry tools
+gem 'pry'
+gem 'pry-stack_explorer'
+gem 'pry-rescue'
+
 # Optional middleman dependencies, included for tests
 gem 'haml', '>= 4.0.5', require: false
 gem 'sassc', '~> 1.8', require: false
--- a/middleman-core/features/page-id.feature
+++ b/middleman-core/features/page-id.feature
@@ -3,14 +3,23 @@ Feature: Page IDs
   Scenario: link_to works with blocks (erb)
     Given the Server is running at "page-id-app"
     When I go to "/index.html"
-    Then I should see "I am: index.html"
+    Then I should see "I am: index"
     And I should see "URL1: /fm.html"
     And I should see "URL2: /2.html"
     And I should see 'URL3: <a href="/3.html">Hi</a>'
     And I should see 'URL4: <a href="/overwrites/from-default.html">Sym</a>'
+    And I should see 'URL5: <a href="/implicit.html">Imp</a>'
+    And I should see 'URL6: <a href="/folder/foldern.html">Foldern</a>'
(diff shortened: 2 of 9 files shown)


tdreyno:
Cool. On Vacation right now, will check in soon.

I do think the extension is important to keep. `index.html` and `index.htm` aren't the same file and shouldn't be treated as a single `index` concept.

joallard:
Thanks! That might be a good point. We'd allow that, two different pages with the same no-extension basename?

I mean, in most of the cases, if you got one `hello.en.html` and one `hola.es.html`, you expect `page_id: hello` to work, don't you?

tdreyno:
Are we already constraining `page_id` to a type (html, xml, etc)? Or is it for all files? Seems like something like `feed.html` and `feed.xml` could happen. Or `header.jpg` and `header.gif`

joallard:
I don't see images or other non-documents assets needing one. The use of a page ID in the first place was to associate different language versions of a document, named differently, with a base ID, possibly written in its frontmatter.

What if, for example, I had a `feed.(en.)xml` and a `fil.fr.xml`? Page IDs could also make sense in that way... Okay, you could extend the argument to images also, if eg. I have two images named differently and a config instruction assigns them their "page" (more like "resource") id. (Though naming image file names in their language/prettifying the names is quite rare in my experience, since an `img[src]` is more of an under-the-hood thing.)

Assuming we want every resource to have a "page_id", I see two ways out:

1. Like I said, make the frontmatter a layer of sugar, and internally append `.html` to every page_id that has no extension.
2. Or, require an extension for every page_id except `.html` (the reverse)

joallard:
Reviewing my comments in #1466, that caught my attention:

> Next up: remove the (*) localizable folder. With I18n activated, we'll assume a page is localizable if it's html, unless it's got `localizable: false` in it.

The "if it's Html" condition sounds awfully familiar.

tdreyno:
@joallard Can we make the "filename to page_id" implementation an option with a Proc that defaults to no extension? I just want to leave a backdoor for people doing strange things, but I think no extension makes sense for 95% of users.

joallard:
Sounds good, I agree people should be able to do strange things.

tdreyno:
@joallard Looks great. I think we could have `remove_extensions` just be another proc and keep the config option as a single thing/type. Maybe rename to `page_id_generator`.

Should try to avoid arity too. Let's just have one way to do a thing and document it well.

joallard:
Good points. Should we feed the path or rather the resource into the proc?

joallard:
Rebased and stuff, we're feeding path to the proc. Rebased, all commits atomic and supposed to be green.

tdreyno:
Awesome. Thanks so much for the hard work! Owe you a 🍺 :)

joallard:
Thanks, feels pretty good! Though to contribute to a project this well maintained and to have that merge are quite rewarding in themselves. I've said it before, but I mean it.

On to #1466 and beyond!

PR #1999, opened 2016-10-27
Fix source watcher configuration

reneklacan:
Problem is that `FileWatcher` is created in `before_sitemap` callback which results in default config all the time, no matter what are user's settings for `watcher_disable`, `watcher_force_polling`, `watcher_latency` and `watcher_wait_for_delay`.

This PR fixes it by configuring `SourceWatcher` instance in `FileWatcher`'s `after_configuration` callback.

### DIFF
--- a/middleman-core/lib/middleman-core/application.rb
+++ b/middleman-core/lib/middleman-core/application.rb
@@ -203,6 +203,7 @@ module Middleman
     define_setting :watcher_disable, false, 'If the Listen watcher should not run'
     define_setting :watcher_force_polling, false, 'If the Listen watcher should run in polling mode'
     define_setting :watcher_latency, nil, 'The Listen watcher latency'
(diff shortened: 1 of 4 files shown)


tdreyno:
This is great, I think I was only using those options from the command line so didn't notice it not working if done via `config.rb`.

One request, can you keep the old `initialize` signature the same. I'd rather not break API compatibility. Even if the `options` is no longer used, let's keep the argument list the same.

Thanks!

tdreyno:
Taking a look now.

tdreyno:
This fixes it: https://github.com/middleman/middleman/commit/b4699cd741f22425b45cb7751f2c358f68ae6bb0

Basically, when a source path changes, re-poll (since the watch is not not running during the change).

reneklacan:
@tdreyno Oh, thank you!

I cherry picked your commit to my branch.

tdreyno:
Thank you!

PR #2006, opened 2016-11-15
Sitemap RedirectResource's know their source file

### LINKED ISSUE #1166: Trouble with Redirects and Sitemap
Hello,

I'm using the [redirects feature](https://github.com/middleman/middleman/blob/v3-stable/middleman-core/lib/middleman-core/sitemap/extensions/redirects.rb) to maintain paths when my Middleman website changes. Unfortunately, I've been experiencing some issues in the development environment.

I made a [test case](https://github.com/ahhrrr/middleman-redirect-test-case) to illustrate my problem. In `config.rb`, I have a redirect set up like so:

```
redirect 'redirect_me/index.html', to: 'destination.html'
```

The redirect seems to work as expected, but when I browse to the Sitemap web view, I get the following error in the console:

```
NoMethodError: super: no superclass method `get_source_file' for #<Middleman::Sitemap::Extensions::Redirects::RedirectResource:0x007fb10ea7d868>
```

And here's what I see in my browser:
screen shot 2014-02-03 at 1 52 32 pm

Am I configuring the redirect incorrectly, or is this a bug? 

Also, I haven't been able to find any documentation, other than the Cucumber feature, on the redirect feature. Is it fully supported? Can I help with documentation?

Thanks!

matiasgarciaisaia:
So Middleman won't hang while rendering the sitemap in development.

I couldn't actually test this - the middleman specs took ages to run, and I haven't found how to create a middleman site that depends on my version of the gem :/ - but monkey-patching this change in my own middleman site worked.

Fixes #1166

### DIFF
--- a/middleman-core/lib/middleman-core/sitemap/extensions/redirects.rb
+++ b/middleman-core/lib/middleman-core/sitemap/extensions/redirects.rb
@@ -68,7 +68,7 @@ module Middleman
 
           # rubocop:disable Style/AccessorMethodName
           def get_source_file
-            nil
+            @request_path.source_file
           end
           # rubocop:enable Style/AccessorMethodName
 


sandstrom:
@matiasgarciaisaia Perhaps this can help with using your own gem:

```ruby
# Gemfile

gem 'middleman', :path => '/path/to/middleman/on/your/computer'

```

Without this change, does it always hang the sitemap in development when a redirect is present, or only under certain circumstances?

matiasgarciaisaia:
@sandstrom I'm not sure how to first pick my local gem, and then run `middleman init` with _that_ local version. Wouldn't that make a change?

The sitemap was useless without this patch if you had a redirection enabled. `Resource#source_file` (what [the Sitemap page](https://github.com/middleman/middleman/blob/29693e4533433238f6f7bd7ffdc0c35ca6445854/middleman-core/lib/middleman-core/meta_pages/sitemap_tree.rb#L42) [asks](https://github.com/middleman/middleman/blob/29693e4533433238f6f7bd7ffdc0c35ca6445854/middleman-core/lib/middleman-core/meta_pages/sitemap_resource.rb#L20) [for](https://github.com/middleman/middleman/blob/29693e4533433238f6f7bd7ffdc0c35ca6445854/middleman-core/lib/middleman-core/meta_pages/sitemap_resource.rb#L42)) is defined as [`@source_file || get_source_file`](https://github.com/middleman/middleman/blob/29693e4533433238f6f7bd7ffdc0c35ca6445854/middleman-core/lib/middleman-core/sitemap/resource.rb#L32-L34), and the only definition of `get_source_file` in the project is [the one on `Middleman::Sitemap::Extensions::Proxies::ResourceInstanceMethods`](https://github.com/middleman/middleman/blob/29693e4533433238f6f7bd7ffdc0c35ca6445854/middleman-core/lib/middleman-core/sitemap/extensions/proxies.rb#L56-L63):

```ruby
          # rubocop:disable AccessorMethodName
          def get_source_file
            if proxy?
              proxied_to_resource.source_file
            else
              super
            end
          end
```

But a `RedirectResource` is not a `proxy?`, so it goes the `super` way to look for a `get_source_file` method... that's not defined.

What I've found, too, is that _it doesn't actually hang_ - it's just that it takes _so long_ to crash (3m24s the whole request), that it looks like it hangs.

But, once again - this fixes it.

matiasgarciaisaia:
I was able to test this locally. Maybe it was a bit redundant, but I did like this:

0 - Edit `/path/to/middleman/`'s `VERSION` constant to 3.4.2 - just for being sure which one I was using
1 - Create an empty dir with a `Gemfile` that only said `gem 'middleman', '3.4.2', path: '/path/to/middleman'`
2 - `cd` to that dir, and `bundle install`
3 - `bundle exec middleman init app`
4 - `cd app && bundle exec middleman` - everything is working
5 - Edited `config.rb` to add a `redirect`:
```
ready do
  redirect 'testing.html', to: sitemap.find_resource_by_destination_path('index.html')
end
```
6 - Wait for reload, test if sitemap works - it does.
7 - Go to `/path/to/middleman` and `git checkout HEAD~1` so the fix doesn't get applied.
8 - Restarted middleman and went to sitemap - it now fails.

Feel free to test or report any issue 👍

matiasgarciaisaia:
Sorry, I'm actually lying.

This #1166 was already fixed in 65d3a0c84f7a42805adbbfa712c24bf34eb54fb0, released ever since v3.3.7. Just returning `nil` didn't work for me, as I didn't have the `Dynamic` label on the sitemap's code - that's why I was returning _some_ path.

What I can't get know is why @milesmatthias was facing the same issue using v3.4.0.

But this PR doesn't have to be merged - the issue was already fixed before.

Sorry to bother you :/

sandstrom:
@matiasgarciaisaia Happy to hear that the issue was resolved! 

And thanks for spending time trying to identify this issue, much appreciated! (even though it turned out to be solved already, your help is still valuable!)

Issue #2017, opened 2016-11-29
Every time I run "middleman" I get a wall of "Expected <> default value for '--<>'"


rr1000:
When I run any middleman command I'm met with this block of text. I've uninstalled and reinstalled several times and can't find a remedy. Any help is greatly appreciated.

```
[0] Expected string default value for '--instrument'; got false (boolean)
[0] Expected boolean default value for '--disable-sitemap'; got :undefined (string)
[0] Expected boolean default value for '--exit-before-ready'; got :undefined (string)
[0] Expected boolean default value for '--https'; got :undefined (string)
[0] Expected boolean default value for '--protect-from-csrf'; got :undefined (string)
[0] Expected boolean default value for '--relative-links'; got :undefined (string)
[0] Expected boolean default value for '--strip-index-file'; got :undefined (string)
[0] Expected boolean default value for '--trailing-slash'; got :undefined (string)
[0] Expected boolean default value for '--watcher-disable'; got :undefined (string)
[0] Expected boolean default value for '--watcher-force-polling'; got :undefined (string)
[0] Expected string default value for '--instrument'; got false (boolean)
[0] Expected boolean default value for '--disable-sitemap'; got :undefined (string)
[0] Expected boolean default value for '--exit-before-ready'; got :undefined (string)
[0] Expected boolean default value for '--https'; got :undefined (string)
[0] Expected boolean default value for '--protect-from-csrf'; got :undefined (string)
[0] Expected boolean default value for '--relative-links'; got :undefined (string)
[0] Expected boolean default value for '--strip-index-file'; got :undefined (string)
[0] Expected boolean default value for '--trailing-slash'; got :undefined (string)
[0] Expected boolean default value for '--watcher-disable'; got :undefined (string)
[0] Expected boolean default value for '--watcher-force-polling'; got :undefined (string)
```

## Steps to reproduce the problem (from a clean middleman installation)
Any middleman command; ex:`$ middleman server`

## Additional information

- Ruby version:  `ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin15]`
- Middleman version: `Middleman 4.1.12`
- OS version: `macOS Sierra 10.12.1`

tdreyno:
Can you try out v4.1.11 and v4.1.10 and see if this was a recent breakage?

iwarner:
@tdreyno Thomas it is - I just upgraded and downgraded and seems its being a little too verbose - still everything functions for me, its just output as far i can tell

[...]

PR #2019, opened 2016-12-03
v4.1.13 Always logging boolean on starting up

5t111111:
In v4.1.13, Middleman always puts logs like `true`/`false` on starting up.
Though I'm not sure if it is intended, it seems to be logs for debugging for the changes for #2017, so I have simply removed that line.


tdreyno:
Whoops, try 4.1.14

PR #2040, opened 2017-01-18
Don't ignore Markdown options in Haml filter

olivierlacan:
This issue was detailed in #1152.

Essentially we're not passing the :markdown options defined in config.rb to
the Markdown renderer when we first go through the Haml renderer.

The solution I've found is to fetch the appropriate options (not just for
Markdown but for Sass or SCSS also) within the Haml renderer's filter loop which
means this patch should also fix any ignored options for Sass and SCSS.

### DIFF
--- a/middleman-core/lib/middleman-core/renderers/haml.rb
+++ b/middleman-core/lib/middleman-core/renderers/haml.rb
@@ -52,6 +52,10 @@ module Middleman
               modified_options = options.dup
               modified_options[:context] = compiler_options[:context]
 
+              renderer_symbol = self.to_s.split(":").last.downcase.to_sym
+              renderer_options = modified_options[:context].app.config[renderer_symbol]
+              modified_options = modified_options.merge(renderer_options || {})
+
               text = template_class.new(nil, 1, modified_options) { text }.render
               super(text, compiler_options)
             end
--- a/middleman-core/features/markdown_redcarpet_in_haml.feature
+++ b/middleman-core/features/markdown_redcarpet_in_haml.feature
@@ -20,6 +20,26 @@ Feature: Markdown support in Haml
     Then I should see ">H1</h1>"
     Then I should see "<p>paragraph</p>"
 
+  Scenario: Markdown filter options in Haml works
+    Given a fixture app "markdown-in-haml-app"
+    And a file named "config.rb" with:
+      """
+      set :markdown_engine, :redcarpet
+      set :markdown, { with_toc_data: true }
+      activate :directory_indexes
+      """
+    And a file named "source/markdown_filter.html.haml" with:
+      """
+      :markdown
+        # Memorable Title
+
+        paragraph
+      """
+    Given the Server is running at "markdown-in-haml-app"
+    When I go to "/markdown_filter/"
+    Then I should see ">Memorable Title</h1>"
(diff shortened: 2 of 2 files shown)


middleman-core/lib/middleman-core/renderers/haml.rb
```diff
@@ -52,6 +52,10 @@ def self.render_with_options(text, compiler_options)
               modified_options = options.dup
               modified_options[:context] = compiler_options[:context]
 
+              renderer_symbol = self.to_s.split(":").last.downcase.to_sym
```
olivierlacan: I agree this is a nasty but I'm fairly unfamiliar with what's available for introspection here and I didn't want to make the patch too Markdown-specific since this is a loop for multiple renderers.

middleman-core/lib/middleman-core/renderers/haml.rb
```diff
@@ -52,6 +52,10 @@ def self.render_with_options(text, compiler_options)
...
+              renderer_options = compiler_options[:context].app.config[renderer_symbol]
```
olivierlacan: I was curious whether accessing the app's config through the context would be frowned upon but I did notice that the Sass renderer itself does something similar: https://github.com/olivierlacan/middleman/blob/9d2cf7b7a0366079958c9e1046d5cd1df8305268/middleman-core/lib/middleman-core/renderers/sass.rb#L90

olivierlacan:
Looks like an unrelated failure in [liquid.feature](https://travis-ci.org/middleman/middleman/jobs/192944775#L7622)

middleman-core/lib/middleman-core/renderers/haml.rb
```diff
@@ -52,6 +52,10 @@ def self.render_with_options(text, compiler_options)
               modified_options = options.dup
               modified_options[:context] = compiler_options[:context]
 
+              renderer_symbol = self.to_s.split(":").last.downcase.to_sym
+              renderer_options = modified_options[:context].app.config[renderer_symbol]
+              modified_options = modified_options.merge(renderer_options || {})
```
olivierlacan: Looks like this won't work in the context of a Haml filter because the `context` is `nil`. If anyone can think of a way to retrieve the `config` some other way, I'll take the hint. 😄

olivierlacan:
I don't think my PR actually quite solves the issue because for some reason I'm getting calls to `render_with_options` which don't include any context, so despite [this self-described hack](https://github.com/middleman/middleman/commit/305d2f99edc0d228e18fe67632f73efe66567ccf#diff-c4f519b4a19b99b4128c9f71635e6e9bR7) I think we're not getting the context which means Haml filters can't read the app configuration for Markdown rendering options.

olivierlacan:
Closing this since the solution is apparently to override the default Markdown filter config in Haml by literally overriding the filter itself. So I can't expect Middleman to do this for me. 

```ruby
module Haml::Filters
  remove_filter("Markdown") #remove the existing Markdown filter

  module Markdown
    include Haml::Filters::Base

    def render(text)
      Redcarpet::Markdown.new(Redcarpet::Render::HTML.new(with_toc_data: true)).render(text)
    end

  end
end
```

marnen:
Yuck.

PR #2049, opened 2017-02-11
clarify callbacks docs

manuelmeurer:
I'm trying to understand how callbacks changed from Middleman v3 to v4. Is it correct that `before_configuration`, `after_configuration`, `before_build`, `after_build`, and `ready` must be defined as separate methods now, and `before_render` and `after_render` are still methods on `app` in the extension`s `initialize` method?


tdreyno:
I believe that is correct. Not sure why the render ones got left behind, maybe just because I assumed they were more internal hooks.

PR #2050, opened 2017-02-11
Remove Compass

tysongach:
This is a WIP, as I don’t really know what I’m doing here.

[Compass] is no longer actively maintained, so we should remove it from Middleman.

[Compass]: https://github.com/Compass/compass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,4 +24,3 @@ Makefile
 .idea
 *.sublime-workspace
 /bin
-middleman-core/fixtures/compass-sprites-app/source/images/icon-s0de2218f58.png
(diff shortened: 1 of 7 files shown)


tdreyno:
Sounds smart to me.

tdreyno:
That said, it's already in an extension so we can keep it for legacy

tysongach:
@tdreyno I’m not sure I’m following. Is the extension fully agnostic and these changes are fine (most are descriptions/mentions of Compass)?

My only hesitation was the changes to the `Gemfile` and `middleman/middleman.gemspec` ( the dependency for `compass-import-once`).

What prompted this issue initially was me being prompted about Compass when doing a `middleman init`. I just realized this is actually coming from the default template: https://github.com/middleman/middleman-templates-default/blob/master/Thorfile#L13-L15

Thoughts on removing that prompt?

tysongach:
I put up a PR to remove the Compass prompt: https://github.com/middleman/middleman-templates-default/pull/9

tysongach:
@tdreyno With the removal of the Compass prompt in the `middleman-templates-default`, I think these changes are still valid, yeah? The changes here seem to be leftover from legacy built-in Compass. People should still be able to use the extension if needed.

Issue #1798, opened 2016-02-09
Localizing paths and links with i18n and directory_indexes (v4.1.1)


sn3p:
This is almost a duplicate of https://github.com/middleman/middleman/issues/1587, from when I was using `v3.4.0`.
Now I'm giving this another try in `v4.1.1`, but still can't get it to work properly ;(

The actual pages are translated correctly, but the `link_to` and `url_for` output is not.
I've created a new project to test this.

**config.rb**

``` rb
activate :directory_indexes
activate :i18n, mount_at_root: :en, langs: [:en, :nl]
```

(also tried activating `i18n` first)

**locales/nl.yml**

``` yml

---
nl:
  paths:
    welcome: welkom

  welcome: Welkom
```

**localizable/index.html.erb**

``` erb
<%= link_to t("welcome"), "/welcome.html" %>
<%= url_for "/welcome.html" %>
```

output (`http://localhost:4567/nl`):

``` html
<a href="/welcome/">Welkom</a>
/welcome/
```

expected:

``` html
<a href="/nl/welkom/">Welkom</a>
/nl/welkom/
```

**Without `directory_indexes` paths are correctly translated unless they have hashes or params**:

```
<%= url_for "/welcome.html" %>
<%= url_for "/welcome.html#foo" %>
<%= url_for "/welcome.html?foo=bar" %>
```

output (`http://localhost:4567/nl`):

``` html
/nl/welkom.html
/welcome.html#hash
/welcome.html?foo=bar
```

(Paths with hashes or params do work when I remove the `paths` translations)

All a bit confusing... Am I overlooking something or is this a bug?

tdreyno:
This looks like a combo of bugs. Thank you for the fantastic test cases. I'll get this fixed up soon.

sn3p:
Thanks, looking forward to it!

For your information, this does work:
```rb
link_to t("menu.pricing"), extensions[:i18n].localized_path("/pricing.html", I18n.locale)
url_for extensions[:i18n].localized_path("/pricing.html", I18n.locale)
```

But breaks when using hashes or params.

slash7:
Looks like still not fixed on v4.1.10.
Any news on this ?

joallard:
Related to #1466

andreamoro:
Not sure I'm late here, but why don't you use the paths: (for the i18n).

Simply add your locale language file, then something like this:

en.yml
\_ paths:
    \___something: "name-of-your-local-page"

it.yml
[...]

Issue #1587, opened 2015-08-26
Path localization with i18n and directory_indexes (v3.4.0)


sn3p:
In v3 `directory_indexes` should be activated AFTER `i18n` (see https://github.com/middleman/middleman/issues/1525).
But then the i18n `url_for` [helper](https://github.com/middleman/middleman/blob/v3-stable/middleman-core/lib/middleman-more/core_extensions/i18n.rb#L63-L88) will not translate paths. When `directory_indexes` is activated BEFORE `i18n` then paths are translated, but then `directory_indexes` isn't working.

Is this a knows issue and is there a workaround?

tdreyno:
Yikes. Let me look

sn3p:
Also the `url_for` [i18n helper](https://github.com/middleman/middleman/blob/v3-stable/middleman-core/lib/middleman-more/core_extensions/i18n.rb#L63-L88) doesn't handle fragments, anchors and query params, like the [core helper](https://github.com/middleman/middleman/blob/f5fe46cb593dcbe57baf870320ca3f8978ca6377/middleman-core/lib/middleman-core/util.rb#L191-L198) does.

```erb
<%= url_for("producten.html", locale: :de) %><br>
<%= url_for("products.html#test", locale: :de) %><br>
<%= url_for("products.html?test=1", locale: :de) %><br>
```
Outputs:
```html
de/produkte.html
products.html#test
products.html?test=1
```

karland:
@tdreyno Any updates to this? Cheers.

tdreyno:
Pretty sure this is almost entirely fixed on v4, other than #1798

andreamoro:
I have been trying to use the url_for, on MiddleMan 4 changing the [examples](https://github.com/middleman/middleman/issues/1587#issuecomment-135469826) so to reflect my locale, but I can't get anything out other than the image path specified. Nothing else is add.

Is the url_for relative or depending by some internal setting like css_dir or images_dir?

PR #2062, opened 2017-03-14
I18n: Keep fragment and query in url_for

juls:
As mentioned in a [comment](https://github.com/middleman/middleman/issues/1587#issuecomment-135469826) on #1587 and in #1798, fragments and queries are discarded when a path is internationalized. The failing test is part of this PR (see 9d5a27ef562921bb2ffa38163389e67d1f1f7031).

This happened because the path _including_ fragment and query was looked up in the internationalization hash. This resulted in an unsuccessful lookup with a fallback to the non-internationalized path. My fix (see 9c1ca5c85af193a32dffe48ebccfc216aa47c047) is to _only_ internationalize the path and leave the rest of the url untouched.

### DIFF
--- a/middleman-core/features/i18n_link_to.feature
+++ b/middleman-core/features/i18n_link_to.feature
@@ -162,6 +162,7 @@ Feature: i18n Paths
       es:
         paths:
           hello: "hola"
+          form: "formulario"
         msg: Hola
       """
     And a file named "source/localizable/hello.html.erb" with:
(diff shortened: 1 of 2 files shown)


tdreyno:
@juls Sorry for taking so long to review this. This is fantastic. Thank you so much!

juls:
@tdreyno no problem, thank you for supporting Middleman. I really appreciate your work for this open source project 🥇

PR #2065, opened 2017-03-21
Add support for locale suffixes to link_to

iblue:
There is a bug in the I18n part which does not link paths including a locale suffix (e.g. `foobar.en.html.erb`) correctly. This patch fixes it and also adds tests.

Previously, `link_to "Foobar", "/foobar.html"` created a link to `foobar.html` regardless of the locale if the file was named `foobar.en.html.erb`. Now it creates the correct links to `/:locale/foobar.html`.

Documentation of the locale suffixes feature can be found at [the "Localizing Entire Templates" section](https://middlemanapp.com/advanced/localization/#localizing-entire-templates), but until now it did not work with `link_to`. For details see the tests.

I hope you like it.


tdreyno:
@iblue Sorry it took so long to review this. This is excellent. Thank you!

Issue #2083, opened 2017-04-24
Build error using liquid


andrezimpel:
I'm using middleman 4.0 with the liquid template engine. Running the middleman server everything is fine and I get no errors. As soon as I want to build the website I get this error:

`Template local 'data' tried to overwrite an existing context value. Please rename the key when passing to 'locals'`.

To be honest I don't know where the error could be since I'm using everything as it shipped, so no extra config except the `set :liquid, :layout_engine => :liquid` in the config.rb. I only got a index.html.liquid file with some includes.

I hope somebody can help me out on this one. Thank you!

iwarner:
@tdreyno @andrezimpel 

Verified this.

Seems a raw liquid usage is producing this error

http://www.rubydoc.info/github/middleman/middleman/Middleman/Renderers/Liquid

```
def manipulate_resource_list(resources)
  return resources unless app.extensions[:data]

  resources.each do |resource|
    next if resource.file_descriptor.nil?
    next unless resource.file_descriptor[:full_path].to_s =~ %r{\.liquid$}

    # Convert data object into a hash for liquid
    resource.add_metadata locals: {
      data: stringify_recursive(app.extensions[:data].data_store.to_h)
    }
  end
end
```

data is reference here

I have tested with and without the data folder and its producing this issue.

iwarner:
@tdreyno 

Just looking into liquid more, definite issue here I think, if you want to include partials from another directory - it will fail

Should be an easy fix, but just wondered if you have some magic for this.

```
# File 'middleman-core/lib/middleman-core/renderers/liquid.rb', line 14

def read_template_file(template_path)
  file = app.files.find(:source, "_#{template_path}.liquid")
  raise ::Liquid::FileSystemError, "No such template '#{template_path}'" unless file
  file.read
end
```

zofrex:
I have run into this and can confirm it happens out of the box.

### Steps to reproduce:

* `middleman init`
* Add 'liquid' to Gemfile and `bundle install`
* Rename `source/index.html.erb` to `source/index.html.liquid`
* `middleman build`

Produces the error:

<details>
<summary>
uncaught throw "Template local `data` tried to overwrite an existing context value. Please rename the key when passing to `locals`"</summary>

```
/Users/zofrex/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/middleman-core-4.3.8/lib/middleman-core/template_renderer.rb:140:in `throw'
[...]

PR #2084, opened 2017-04-25
allow to override data in context

andrezimpel:
Currently the build fails when liquid tries to override the (empty or not existing?) context value data.

cc: https://github.com/middleman/middleman/issues/2083

### DIFF
--- a/middleman-core/lib/middleman-core/template_renderer.rb
+++ b/middleman-core/lib/middleman-core/template_renderer.rb
@@ -132,7 +132,7 @@ module Middleman
       @app.extensions.add_exposed_to_context(context)
 
       locals.each do |k, _|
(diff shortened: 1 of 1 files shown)


tdreyno:
So will this allow Liquid to overwrite Middleman's `data` provider? Or just allow Liquid templates to access it?

andrezimpel:
It will allow Liquid to override the empty data context while rendering. 

`Template local 'data' tried to overwrite an existing context value. Please rename the key when passing to 'locals'` the context value has been empty or is not even existing in mutplile scenarios I have.

tdreyno:
@andrezimpel Do you have YAML files in the `data` folder? If not, can you test with that?

andrezimpel:
I did also test that and everything works fine. What's the intention of context data anyways?

PR #2099, opened 2017-06-20
Fix for liquid partial include - and passing tests

iwarner:


tdreyno:
Is this second commit on purpose? Seems unrelated?

iwarner:
Ahh man - so if I commit to my master then it pulls in the whole branch not the atomic push request. Still learning this.

Can you validate that the liquid one is okay? and I will recreate via branches on my fork

PR #2107, opened 2017-08-13
Processor lanes

dtorres:
The idea of this PR is to reduce the amounts of times the mutex is hit to just the number of spawned processes.

In the current implementation, the mutex is hit every time a resource is rendered, which given a lengthy job is not that important, however many websites have multiple resources of insignificant render time.

The potential downside of this is uneven processing when a portion of the resources are more expensive than the rest.

My website is too small (~100 resources) to give me a significant metric, however I created a test of 1000 html.erb files and it seemed to save a couple of seconds off the total time

I would like to ask if some could try this patch and report back their findings.

Thanks.

TODO:
- [ ] Rebase/squash the commits

### DIFF
--- a/middleman-core/lib/middleman-core/builder.rb
+++ b/middleman-core/lib/middleman-core/builder.rb
@@ -134,12 +134,14 @@ module Middleman
     Contract ResourceList => ResourceList
     def output_resources(resources)
       results = if @parallel
-        threads = 4
-        min_parts = resources.count/threads
-        reminder = resources.count%threads
+        res_count = resources.count
+        processes = ::Parallel.processor_count
+        processes = processes < res_count ? processes : res_count
+        min_parts = res_count/processes
+        reminder = res_count%processes
         offset = 0
         ranges = []
-        while offset < resources.count
+        while offset < res_count
           end_r = offset + min_parts
           if reminder > 0
             end_r +=1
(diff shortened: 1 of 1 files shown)


middleman-core/lib/middleman-core/builder.rb
```diff
@@ -134,7 +134,29 @@ def output_files
...
+          if reminder > 0
+            end_r +=1
+            reminder -= 1
+          end
+          range = offset...end_r
+          offset = end_r
+          ranges << range
+        end
+        outputs = Parallel.map(ranges, in_processes: processes) do |range|
+          puts range.to_s
```
tdreyno: Leftover log?

middleman-core/lib/middleman-core/builder.rb
```diff
@@ -134,7 +134,29 @@ def output_files
     Contract ResourceList => ResourceList
     def output_resources(resources)
       results = if @parallel
-        ::Parallel.map(resources, &method(:output_resource))
+        res_count = resources.count
+        processes = ::Parallel.processor_count
+        processes = processes < res_count ? processes : res_count
+        min_parts = res_count/processes
+        reminder = res_count%processes
```
tdreyno: Typo. "Remainder" ?

PR #2116, opened 2017-09-20
Use i18n fallbacks when looking up localized paths

schmidt:
In #853, the usage of i18n-fallbacks was introduced. For undocumented reasons, the fallbacks are ignored for localized paths, though.

This commit removes this limitation, adds a test to make sure, that it works as expected, and extends fixtures to make sure, that the other tests work with the changed behavior.


schmidt:
I've updated the MR with the following objectives:

* Provide a better example in the test data
* Separate feature and fixing of unrelated tests

PR #2128, opened 2017-11-15
remove fast_blank dependency

mojavelinux:
- remove fast_blank as it blocks Middleman from being used with JRuby


mojavelinux:
I don't see any references to fast_blank in the code base. If there are, and we want to support JRuby, we need to make fast_blank an optional dependency and fall back to something more compatible.

It would also be nice if middleman was tested on JRuby to avoid these types of incompatibilities from creeping in in the future.

mojavelinux:
Here's proof I can use middleman with JRuby without this dependency (which is blocking installation):

https://travis-ci.org/middleman/middleman-asciidoc/jobs/302338616

PR #2131, opened 2017-12-01
Loosen gem requirement on I18n

johnnyshields:


tdreyno:
Is this blocking something?

johnnyshields:
I came across this when I was trying to add rails and middleman in the same ruby project.

PR #2132, opened 2017-12-02
clear lazy map after resolving Tilt templates

mojavelinux:
- clear lazy map after resolving Tilt templates
- use raw ext to resolve Tilt template

### DIFF
--- a/middleman-core/lib/middleman-core/application.rb
+++ b/middleman-core/lib/middleman-core/application.rb
@@ -345,13 +345,14 @@ module Middleman
 
     # Clean up missing Tilt exts
     def prune_tilt_templates!
-      ::Tilt.default_mapping.lazy_map.each_key do |key|
+      mapping = ::Tilt.default_mapping
+      mapping.lazy_map.each_key do |key|
         begin
-          ::Tilt[".#{key}"]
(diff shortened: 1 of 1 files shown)


mojavelinux:
Since this method resolves the Tilt template for each extension, there's no reason to keep entries in the lazy map. If entries are left in the lazy map, this can later lead to the extensions_for method returning duplicate entries.

I also simplified the call to resolve the template using `mapping[ext]` instead of `::Tilt[".#{ext}"]` since these two calls do exactly the same thing.

mojavelinux:
This may not be necessary if Tilt accepts the change upstream. See https://github.com/rtomayko/tilt/pull/324.

One way or another, we need agreement on this.

tdreyno:
@mojavelinux Are we still waiting on that upstream patch for this?

mojavelinux:
Still outstanding. See https://github.com/rtomayko/tilt/pull/324

Do you need me to rebase this?

tdreyno:
Nope, I'll merge this fix for now

mojavelinux:
:+1:

mojavelinux:
:+1:

PR #2140, opened 2018-01-16
Fix redirect destination lookup

### LINKED ISSUE #2011: Redirect issue

Redirect example from the doc is not working: 
```
redirect "/my/old/path.html", to: "/my/new/path.html"
```

but

```
redirect "my/old/path.html", to: "/my/new/path.html"
```

is working fine.

I dig a bit into the issue. 

It seems to come from here https://github.com/middleman/middleman/blob/7c968b9572923fd0f7fb176ca8e52a91727adf84/middleman-core/lib/middleman-core/sitemap/store.rb#L155
`request_path` is `my/old/path.html` but the key in `@ _lookup_by_destination_path` is `/my/old/path.html`

## Additional information

- Ruby version: ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin16]
- Middleman version: 4.1.10

matiasgarciaisaia:
A bit late, but this should do the trick 🎉

Fixes #2011

### DIFF
--- a/middleman-core/lib/middleman-core/sitemap/store.rb
+++ b/middleman-core/lib/middleman-core/sitemap/store.rb
@@ -238,11 +238,11 @@ module Middleman
 
                 # Rebuild cache
                 @resources.each do |resource|
-                  @_lookup_by_path[resource.path] = resource
+                  @_lookup_by_path[::Middleman::Util.normalize_path(resource.path)] = resource
                 end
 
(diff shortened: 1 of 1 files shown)


matiasgarciaisaia:
The specs are failing - but I'm not sure it has anything to do with this PR.

Any ideas, @tdreyno?

PR #2143, opened 2018-01-18
Fix #2039 - ignore of I18n files

### LINKED ISSUE #2039: Middleman builds all localized files
I have a multilingual middleman site and each one is deployed on a separate domain. So my I18n extension is setup like this: `activate :i18n, :langs => [ TARGETTED_LOCALE ]`

Even if only one locale is specified, it will build and deploy all localized files, for example:
/localizable/privacy.de.html
/localizable/privacy.es.html
...

I also use a filter for the blog to only build the relevant blog articles this way:
```
activate :blog do |blog|
  blog.filter = proc { |article|
    article.tags.include? target.to_s
  }
end
```
Although the filter is working well, it also looks like it's building all blog articles, not just the filtered ones.

johnnyshields:
Fixes #2039


tai-flywire:
Sorry to be a newb but what is the process for getting this into a release?  It seems like it was in all the development versions of 4.3.0 but didn't make the cut for the final release and every point release since then.  Has it just been forgotten?  It would be really useful for us to have this.

tdreyno:
4.3 became 5.0 which isn't out yet. I can backport stable releases.

tdreyno:
Published as `v4.3.4`

tai-flywire:
Fantastic, thank you!!

PR #2156, opened 2018-02-05
add ExtensionManager#active? to check if extension is active

mojavelinux:
Add a lightweight method to check if an extension is active:

```
if app.extensions.active? :google_analytics
```

I often need this in templates. The alternative is to retrieve the extension:

```
if app.extensions[:google_analytics]
```

but this does not clearly communicate intent.


tdreyno:
Nice. Working on getting some sticky broken tests working and then will get a release out with all this stuff.

PR #2157, opened 2018-02-07
discover template in local directory if applicable

mojavelinux:
- look for template in local directory
- use looser check for SSH git URL

### DIFF
--- a/middleman-cli/lib/middleman-cli/init.rb
+++ b/middleman-cli/lib/middleman-cli/init.rb
@@ -47,7 +47,7 @@ module Middleman::Cli
 
         begin
           data = ::JSON.parse(uri.read)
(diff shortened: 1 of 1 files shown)


mojavelinux:
I wonder if we should require that the local directory be an absolute path (using Pathname(repo_name).absolute?). That might make it less likely to accidentally find a matching local path. wdyt?

I also split out the clone operation to a dedicated function.

tdreyno:
@mojavelinux I think it's a good idea to make it absolute and error loudly in case of conflict.

mojavelinux:
In that case, let's do it!

mojavelinux:
I'll rebase.

mojavelinux:
Rebase complete!

PR #2158, opened 2018-02-08
resolves #2154 allow bundle path to be specified in init command

### LINKED ISSUE #2154: Allow a bundle path to be passed to the init command

Currently, the init command allows the call to bundle to be disabled. However, it would be nice if would be possible to instead pass a path to bundle (or a default project-relative path). This allows the gems to be installed within the project.

(My preference happens to be `--path=.bundle/gems`).

Run the init command without the `-B` flag.

```
middleman init
```

Bundler is run, but installs the gems globally. For some, the desired behavior is to install the gems in the project.

## Additional information

- Ruby version: 2.4.2
- Middleman version: 4.2.1
- OS version: Fedora 26

mojavelinux:


mojavelinux:
:+1:

PR #2159, opened 2018-02-08
better error message when init fails to clone git repo

mojavelinux:


mojavelinux:
Note that the git_path variable could be used in the git command. But then I realized that it's not really the right value for the message. So I simply reworked the message.

tdreyno:
Thanks for all these. Just got the tests passing (3rd party deps changed). Will make it safe to push out a new release.

mojavelinux:
:tada:

PR #2183, opened 2018-08-07
Do not create unnecessary array in IgnoreDescriptor

TheSmartnik:
Hi,
I was investigating some memory problems in my middleman project and found this piece of code. As `ignore!` mutates original object, seems unnecessary to create another array with `map`, `each` will do the same thing here without extra allocation


tdreyno:
Looks good. Will merge after we fix Travis

PR #2188, opened 2018-09-05
Fix localisation and recursion issues of Traversal::parent.

SnijderC:
I'm trying to make breadcrumbs for our website. We are using localisation and have it up like this:

``` ruby
locales = [:en, :nl]
I18n.available_locales = locales
set :locale_paths, :en => "/", :nl => "/nl/"
activate :i18n do |i18n|
  i18n.templates_dir = "pages"
  i18n.locales = locales
end
```

We also have paths that are translated by translation strings in the locale files. E.g.:

``` yaml
paths: 
  your-account: uw-account
```

When `parent` is called on a localised Resource (`current_page`) it returns `nil`, but not for the language mounted at root (`en`). So we have an English page with perfectly rendered breadcrumbs and a Dutch version with breadcrumbs that only contains a link to `current_page`.

I have spend some time trying to debug this and have come to the conclusion that this line is the culprit:

https://github.com/middleman/middleman/blob/3eebc09338b376e0ae51cb8c0de09e1e29a5dbf2/middleman-core/lib/middleman-core/sitemap/extensions/traversal.rb#L18

When I remove this line and change `root` to `path` on the next: 

https://github.com/middleman/middleman/blob/3eebc09338b376e0ae51cb8c0de09e1e29a5dbf2/middleman-core/lib/middleman-core/sitemap/extensions/traversal.rb#L19

The breadcrumbs will show from the `current_page` all the way up to root. However we get root twice, once localised (`/nl/`) and once non-localised (`/`). Seems to me this is inherent to the fact `path` is used to find parents, `/` will always be a parent for `/nl/`. You could argue that this is expected behaviour because you mounted on something on `/`. But I imagine users would not expect it.

I found these related issues by git blaming line 18: 

- https://github.com/middleman/middleman/issues/1709
- https://github.com/middleman/middleman/issues/1733

Which obviously describe the situation before line 18 was added. And also the, apparently not new - problem I face.

My proposed solution is to not return the root if it is equal to `traversal_root` to prevent the duplicate root situation.

Next I identified that when you have this structure: 

``` yaml
- sub1
  - sub2
    - index.html
- index.html
```
The parent of `sub1/sub2/index.html` can not be determined. This is not what you want when you are creating breadcrumbs so I added an optional argument `recurse` to `#parent` that recurses over the next level up, next level up, etc. until an ancestor is found or root is reached.

I also added some lines to the feature file and added the necessary files to the fixture in `i18n-mixed-sources`.

Lastly I added a whole bunch of comments because this was a pita to debug, hopefully whoever needs to make changes or fix bugs next won't have to go through this process.

``` bash
bundle exec cucumber features/i18n_mixed_sources.feature
[snip]
1 scenario (1 passed)
33 steps (33 passed)
```
The other tests that didn't fail before, don't fail after.

Please note, if it isn't obvious from the PR 😄, I do not use Ruby on a daily basis, so  please review carefully.


tdreyno:
Thank you for digging @SnijderC, I'll review this shortly

Issue #2186, opened 2018-08-24
files.watch does not reload pages if a destination_dir is specified


mlimonczenko:
## Overview

We have a middleman project with `files.watch` using a `destination_dir` in the `config.rb`.

**When we do the following:**
1. Run `middleman serve`.
1. See the page load.
1. Edit a file in the `destination_dir` path.
1. Reload the page.

**We get the following result:**
- The shell output indicates that it detects the `destination_dir` file.
- We don't see the updates we made in the local web server.

## Additional Investigation

**When we do the following:**
1. Run `middleman serve`.
1. Make an edit to a file that is in the `/source` directory.

**We get this result:**
- The shell output does not indicate that middleman detects the `/source` file.
- We also do not see it updated on the local webserver.

**But, when we do this:**
1. Run `middleman serve`.
1. Edit a file in the `source` directory.
1. Edit a file in the `destination_dir` directory.
1. Refresh `destination_dir` page and `source` page.

**We get these results:**
- The shell output does not indicate that middleman detects the `/source` file.
- The shell output indicates that it detects the `destination_dir` file.
- We don't see the updates in the `destination_dir` page.
- We see updates to the `source` page.

When we don't specify a `destination_dir` everything works as intended.

## Steps to Reproduce

1. Add `files.watch` with a `destination_dir` to your middleman project `config.rb`.

    ```
    files.watch(
      :source,
      path: File.expand_path("/absolute/path/to/content/directory"),
      destination_dir: "content-directory",
    )
    ```
1. And you do the following:
    1. Run `middleman serve`.
    1. Go to `localhost:4567/content-directory/page.html` and see that it is rendered properly.
    1. Edit the file `/absolute/path/to/content/directory/page.html.md`.
    1. Revisit `localhost:4567/content-directory/page.html`.

**Expected behavior:** You see the updates reflected on the page

**Actual behavior:** You see the page did not change.

## Assistance

I would be happy to help investigate this bug and assist in any way.
Let me know how I can be accommodating.

PR #2192, opened 2018-09-10
Fix failing feature specs

motevets:
Hello middleman team,

@mlimonczenko and I are very interested in getting #2190 pulled in and released in a patch of middleman.  The bug it fixes, #2186, is blocking us from releasing a new version of an open source gem we maintain, [bookbinder](https://github.com/pivotal-cf/bookbinder), of which middleman is a core dependency. We want to help anyway that we can to assist getting our fix pulled in, so we took a crack at fixing the build.  Let us know if there's anything else we can do.

Thank you for building maintaining middleman.  We get a lot out of it, and we hope this helps.


tdreyno:
Thank you @motevets and @mlimonczenko! These hash tests depend on other dependencies and are very brittle.

PR #2202, opened 2018-10-18
nokogiri update due to CVE

altblanc:
see here: https://github.com/sparklemotion/nokogiri/blob/master/CHANGELOG.md#security-notes


tdreyno:
Can you update the the source of this dependency or add the peer requirement of >= the insecure version? It'll probably be in `Gemfile`

PR #2212, opened 2018-11-06
Address security vulnerability due to i18n 0.7.0

cchawn:
### Why
The previous version of i18n has a critical vulnerability that has been addressed in a subsequent release.

```
Name: i18n
Version: 0.7.0
Advisory: CVE-2014-10077
Criticality: Unknown
URL: https://github.com/svenfuchs/i18n/pull/289
Title: i18n Gem for Ruby lib/i18n/core_ext/hash.rb Hash#slice() Function Hash Handling DoS
Solution: upgrade to >= 0.8.0
```

### What
This PR updates i18n to 0.8.0 to address this security vulnerability.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -19,7 +19,7 @@ PATH
       fastimage (~> 2.0)
       hamster (~> 3.0)
       hashie (~> 3.4)
-      i18n (~> 0.7)
+      i18n (~> 0.8)
       listen (~> 3.0)
       memoist (~> 0.14)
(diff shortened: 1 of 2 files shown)


cchawn:
Unable to update to this version of i18n since not all checks pass when running with ruby 2.3.7 😞

tdreyno:
I've been keeping an eye on this, but, yeah, there are issues with the tests when upgraded.

That said, since Middleman is only run locally, this vector shouldn't be exploitable

Issue #2217, opened 2018-11-08
i18n security vulnerability


Ffoeg311:
## Expected behavior and actual behavior
In our implementation of middleman, we also use bundler audit to find security vulnerabilities in CI. Starting on November 6, we started to get this alert:
```
ruby-advisory-db: 328 advisories
Name: i18n
Version: 0.7.0
Advisory: CVE-2014-10077
Criticality: Unknown
URL: https://github.com/svenfuchs/i18n/pull/289
Title: i18n Gem for Ruby lib/i18n/core_ext/hash.rb Hash#slice() Function Hash Handling DoS
Solution: upgrade to >= 0.8.0
```
This is because middleman-core uses i18n 0.7.0 as a dependency. From our Gemfile.lock:
```
middleman-core (4.2.1)
      activesupport (>= 4.2, < 5.1)
      addressable (~> 2.3)
      backports (~> 3.6)
      bundler (~> 1.1)
      contracts (~> 0.13.0)
      dotenv
      erubis
      execjs (~> 2.0)
      fast_blank
      fastimage (~> 2.0)
      hamster (~> 3.0)
      hashie (~> 3.4)
      i18n (~> 0.7.0)
      listen (~> 3.0.0)
      memoist (~> 0.14)
      padrino-helpers (~> 0.13.0)
      parallel
      rack (>= 1.4.5, < 3)
      sass (>= 3.4)
      servolux
      tilt (~> 2.0)
      uglifier (~> 3.0)
```
Updating middleman-core to 0.8.0 should fix this.
## Steps to reproduce the problem (from a clean middleman installation)
Running bundle install, you should see `i18n (~> 0.7)` in your Gemfile.lock.
## Additional information

- Ruby version: 2.3.3
- Middleman version: 4.2.1
- OS version: Mac OS High Sierra

joevus:
Also affecting my work.

tdreyno:
As `i18n` only runs during dev mode, and never in production (which is static), you should be able to safely ignore this.

The library has issues with older Rubies we still support and can't be upgraded cleanly at this time.

PR #2218, opened 2018-11-08
bumped i18n version

Ffoeg311:
This is to fix the issue that I opened here:
https://github.com/middleman/middleman/issues/2217
Just bumping i18n version to fix the security vulnerability in 0.7


tdreyno:
Weird. Wonder why https://github.com/middleman/middleman/pull/2212 failed.

tdreyno:
Going to back port this to the stable series as well

PR #2227, opened 2018-12-15
Remove bundler version constraint

colby-swandale:
Hey! 👋 

I'm the release manager for Bundler and I'm preparing for the upcoming Bundler 2 release. I noticed that `middleman-core` has Bundler locked to `~> 1.1`. This should be removed so that people who only have Bundler 2 installed can install your gem.

You don't have to worry about making sure your `Gemfile.lock` loads with the correct version of Bundler, it will take care of that itself.

Thanks!

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,7 +11,7 @@ PATH
       activesupport (>= 4.2, < 5.2)
       addressable (~> 2.3)
       backports (~> 3.11)
-      bundler (~> 1.1)
+      bundler
       contracts (~> 0.16.0)
       dotenv
       erubis
--- a/middleman-core/middleman-core.gemspec
+++ b/middleman-core/middleman-core.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.3.0'
 
   # Core
-  s.add_dependency('bundler', ['~> 1.1'])
+  s.add_dependency('bundler')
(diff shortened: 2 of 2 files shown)


tdreyno:
Thank you @colby-swandale

purrloftruth:
A Middleman 4.2.2 should be pushed with this change because, now that Bundler 2 is out, those installing Middleman for the first time on a new Ruby installation are unable to use `middleman new` at all:

```
[snip]
Resolving dependencies...
Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    middleman (~> 4.2) was resolved to 4.3.0.rc.1, which depends on
      middleman-core (= 4.3.0.rc.1) was resolved to 4.3.0.rc.1, which depends on
        bundler (~> 1.1)

  Current Bundler version:
    bundler (2.0.1)
This Gemfile requires a different version of Bundler.
Perhaps you need to update Bundler by running `gem install bundler`?

Could not find gem 'bundler (~> 1.1)', which is required by gem 'middleman-core (= 4.3.0.rc.1)', in any of the sources.

Bundler could not find compatible versions for gem "middleman-core":
  In Gemfile:
    middleman (~> 4.2) was resolved to 4.3.0.rc.1, which depends on
      middleman-core (= 4.3.0.rc.1)

    middleman-autoprefixer (~> 2.7) was resolved to 2.9.0, which depends on
      middleman-core (>= 3.3.3)
```

PR #2249, opened 2019-01-26
Feature: Add an option to create asset hashes without using the original filenames as prefix

iblue:
Hi. So I added a feature I would like to have. Using the `remove_filename` option on `asset_hash` now creates assets without prefixing the filenames.

With `remove_filename: false` (default): `javascripts/application-4553338c.js`
With `remove_filename: true`: `javascripts/4553338c.js`

If you need any changes, tests or additional documentation, please just comment. Thank you for this great piece of software!

### DIFF
--- a/middleman-core/features/asset_hash.feature
+++ b/middleman-core/features/asset_hash.feature
@@ -306,3 +306,17 @@ Feature: Assets get file hashes appended to them and references to them are upda
       | javascripts/application.js.map |
 
     And the file "javascripts/application-myprefix-4553338c.js" should contain "//# sourceMappingURL=application.js-myprefix-22cc2b5f.map"
+
+  Scenario: Filenames can be removed by option
+    Given a successfully built app at "asset-hash-remove-filename"
+    When I cd to "build"
+    Then the following files should exist:
+      | index.html |
+      | javascripts/4553338c.js |
+      | javascripts/22cc2b5f.map |
+      | index.html |
+    And the following files should not exist:
+      | javascripts/application.js |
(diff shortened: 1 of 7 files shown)


tdreyno:
Seems reasonable, but how about we go all the way and make this configurable with a Proc/block

```
activate :asset_hash, rename_proc: -> (path, file_name, hash, extension) { "#{path}/#{hash}/{#extension}" }
```

The people can do whatever they want (use less of the hash, put hashes in the path instead of the file name, whatever). We can reimplement the current behavior as the default value of this callback.

iblue:
Sounds good. Is breaking backwards compatibilty an option? If yes, I would also remove the `prefix:` option, which can be handled with the proc as well.

tdreyno:
Hmm, the proc could have a scope which allows it to read `prefix`? I'd like to avoid breaking compat.

iblue:
I implemented the feature as proc, but I think your test suite may be broken. I did not add any tests and I did not remove the existing ones, but the test suite completes successfully - on the CI and on my machine. Any suggestions?

tdreyno:
There's a really odd test that sometimes fails trying to bind an IP on Travis.

PR looks great. Final question, do you want this available in the stable 4.x series (branch `4.x`) or only in the 5.0 betas (`master`)?

tdreyno:
No, I can cherry-pick it. Thanks. I'll merge this tomorrow

iblue:
Awesome, thanks! I quickly fixed the rubocop issues and the build should be green now.

PR #2263, opened 2019-03-28
Reimplement the i18n lookup table

iblue:
I have found middleman to be quite fragile when working with i18n pages. For example `link_to "This page in German", current_page, locale: :de` often just links to the same English version, for example with i18n paths.

This is due to a buggy implementation of the lookup tables used for the paths. I debugged the issue, added tests and a fix that should prove to be more robust in the field. It also seems to be faster (on my machine).


iblue:
BTW: I also fixed a small bug that causes issues with files than contain `.<locale>.` outside of the `localizable/` directory (we use `mail-at-domain.de.asc.key` for our PGP keys). No test included for that one however.

tdreyno:
Excellent. I will review as soon as possible

PR #2274, opened 2019-04-18
Disable pessimistic partial update

### LINKED ISSUE #2273: On changed file the source watcher loads all files with the same extension
If file is changed, especially .haml file the middleman --instrument will show "possible change" also for all other haml-files. This makes reloading really slow. If livereload is used a change in .haml-file can take minutes for livereload to refresh. 

I removed "valid_updates |= related_updates" line from sourcer_wather.rb and made local gem and now it seems lightning fast. 

Is there some specific reason that all same type files should be marked as changed?

The 3.x middleman haml livereload is seconds and default 4.x livereload is minutes. 

## Additional information

- Ruby version: 2.6
- Middleman version: 4.3.3
- OS version: Mac&Linux

tdreyno:
Currently, when a file which could be included as a partial in another file changes (almost all scss, erb and haml), we attempt to "reload" every single file of that type "just in case." This is a result of those templating engines not having dependencies trees we can easily track (in 4.x).

Closes #2273 

Usage:

```
set :pessimistic_partial_update, false
```

### DIFF
--- a/middleman-core/lib/middleman-core/core_extensions/file_watcher.rb
+++ b/middleman-core/lib/middleman-core/core_extensions/file_watcher.rb
@@ -9,6 +9,9 @@ module Middleman
       Contract IsA['Middleman::Sources']
       attr_reader :sources
 
+      # Defaults to true for 4.x, could become false for 5.x
+      define_setting :pessimistic_partial_update, true, 'Whether a change to a partial should cause a pessimistic reload of all files which could include it.'
+
       # Make the internal `sources` method available as `app.files`
       expose_to_application files: :sources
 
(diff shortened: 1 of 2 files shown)


tkarsist:
I took your branch, made local gem but didn't manage to get this working. I had the following in the config.rb:
"set :pessimistic_partial_update, false"

For some reason the setting value is not going through. The Instrument is still printing the "Possible change" which is currently inside the loop and shouldn't be printed.

tdreyno:
You can also have bundler use the repo directly without making a gem:

```
gem "middleman", github: "middleman/middleman, branch: "feature/pessimistic-partial-update"
```

tkarsist:
Thank you. I got also the bundler use of repo working. I'm still unable to disable the partial updates. I have tried the following in config.rb but no success:
set :pessimistic_partial_update, false

It is still loading all of the assets.

PR #2275, opened 2019-05-14
Remove test files from gemspec

TheSmartnik:
Middleman includes specs and fixtures in a gem.
It seems they don't really serve any useful purpose https://github.com/rubygems/rubygems/issues/735 and for example excluded from [bundler generated gems](https://github.com/bundler/bundler/blob/master/lib/bundler/templates/newgem/newgem.gemspec.tt#L24-L26).

If excluded they significantly reduce size of the gems
Before
Screen Shot 2019-05-14 at 20 47 51
Screen Shot 2019-05-14 at 20 58 52
After
изображение
Screen Shot 2019-05-14 at 20 58 47

### DIFF
--- a/middleman-cli/lib/middleman-cli/templates/extension/gemspec
+++ b/middleman-cli/lib/middleman-cli/templates/extension/gemspec
@@ -11,8 +11,7 @@ Gem::Specification.new do |s|
   # s.summary     = %q{A short summary of your extension}
   # s.description = %q{A longer description of your extension}
 
-  s.files         = `git ls-files`.split("\n")
-  s.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n")
+  s.files       = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(fixtures|features|spec)/}) }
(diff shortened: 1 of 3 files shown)


middleman-cli/lib/middleman-cli/templates/extension/gemspec
```diff
@@ -11,8 +11,7 @@ Gem::Specification.new do |s|
   # s.summary     = %q{A short summary of your extension}
   # s.description = %q{A longer description of your extension}
 
-  s.files         = `git ls-files`.split("\n")
-  s.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n")
+  s.files       = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(fixtures|features|spec)/}) }
```
TheSmartnik: It wouldn't work if gems are built from a parent folder. Not sure, if that's ok.

tdreyno: We build gems with the `bundle exec rake publish` command from the root.

TheSmartnik: @tdreyno thanks! Just tested it, works correctly

PR #2278, opened 2019-06-09
Fix invalid Haml option warnings

k0kubun:
When I use middleman-blog with Haml 5+, I get the following warnings

```
Haml::TempleEngine: Option :lang is invalid
Haml::TempleEngine: Option :locale is invalid
```

because middleman-blog adds `lang` and `locale` to metadata:
https://github.com/middleman/middleman-blog/blob/v4.0.2/lib/middleman-blog/blog_data.rb#L235

This patch is for resolving these warnings in the most conservative way. We could opt-in limited keys from `@options`, but taking this conservative way would allow us to avoid unwanted impacts.


tdreyno:
Wondering if we have a test for i18n helpers in haml. I thought these were included for that support.

k0kubun:
Maybe this https://github.com/middleman/middleman/commit/4d5c50968? According to the commit and https://github.com/middleman/middleman/commit/28a0071be, the `options[:locale]` and `options[:lang]` are for controlling `::I18n.locale` and they seem to have nothing to do with any of Haml options.

For your reference, in `Haml::Engine` (and underlying `Haml::TempleEngine`) does not have neither `:locale` nor `:lang` in its options https://github.com/haml/haml/blob/v5.1.1/lib/haml/temple_engine.rb#L10-L30.

k0kubun:
Since you closed this, I guess you removed `:locale` and `:lang` from `@options` in another change? These are still invalid options in the Haml side and continue to be warned if they haven't been.

Issue #2262, opened 2019-03-27
deps.yml is not written when --only-changed is set


smcgivern:
## Expected behavior and actual behavior

In https://github.com/middleman/middleman/issues/2228#issuecomment-475685216, I said:

> Can we clarify what flags people should use in CI? Ideally I'd just run `--only-changed --track-dependencies` on every single build, and use the presence or otherwise of `deps.yml` to tell Middleman whether it's a full build or not

Running `middleman build --track-dependencies --only-changed` on the Middleman master branch does not write `deps.yml`.

Running `middleman build --track-dependencies` does write it.

It would be convenient to have one set of options that will do either. Otherwise we have to update our build scripts to check for `deps.yml` and set `--only-changed` depending on whether it exists or not.

tdreyno:
Sounds reasonable. I'll make `--only-changed` essentially not exist if `deps.yml` doesn't exist yet.

smcgivern:
@tdreyno is there any way I can help with this?

tdreyno:
This is the block which handles that logic: https://github.com/middleman/middleman/blob/master/middleman-core/lib/middleman-core/builder.rb#L58-L90

smcgivern:
Thanks. It seems the issue is: https://github.com/middleman/middleman/blob/master/middleman-core/lib/middleman-core/builder.rb#L125-L127

This deliberately doesn't write the file if no vertices were invalidated, but in the case where the file didn't exist initially, we kind of want to create the file anyway. I've had a bash at that in https://github.com/smcgivern/middleman/commit/840f7f3ff2b3c3c37307a763845605a554556aab but I've only tested it for my case so far.

tdreyno:
I think that makes sense

PR #2279, opened 2019-06-11
Create deps.yml when --only-changed is set

smcgivern:
This allows you to always run with --track-dependencies --only-changed,
and if deps.yml is missing, it will write a new one.

It also fixes a small issue where other types of failure in loading
deps.yml were always treated as a general corruption, because the errors
inherited from StandardError through RuntimeError, so they were
re-raised as the less-specific InvalidDepsYAML.

Closes https://github.com/middleman/middleman/issues/2262.

### DIFF
--- a/middleman-core/lib/middleman-core/builder.rb
+++ b/middleman-core/lib/middleman-core/builder.rb
@@ -54,14 +54,19 @@ module Middleman
     def run!
       @has_error = false
       @events = {}
+      create_deps_yml = false
 
       if @track_dependencies
         begin
           ::Middleman::Util.instrument 'dependencies.load_and_deserialize' do
             @graph = ::Middleman::Dependencies.load_and_deserialize(@app)
           end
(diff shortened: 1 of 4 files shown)


middleman-core/lib/middleman-core/builder.rb
```diff
@@ -124,7 +129,7 @@ def run!
       unless @has_error
         partial_update_with_no_changes = (@only_changed || @missing_and_changed) && @invalidated_files.empty?
 
-        ::Middleman::Dependencies.serialize_and_save(@app, @graph) if @track_dependencies && !partial_update_with_no_changes
+        ::Middleman::Dependencies.serialize_and_save(@app, @graph) if @track_dependencies && (create_deps_yml || !partial_update_with_no_changes)
```
smcgivern: I bailed on adding specs here because `run!` is really long. I tested it manually and it works though.

smcgivern:
@tdreyno could you take a look, please?

PR #2294, opened 2019-10-09
Conform generated gemspec to `bundle gem` style

unasuke:
## Motivation
In executing `$ bundle gem`, gemspec contains git `user.name` and `user.email` automatically. 
And place `TODO` to the field to must update. It's very useful.

## Changes
- remove obsoleteness magic comment
    - 	`# -*- encoding: utf-8 -*-`
- authors and email from git config
- uncomment and add `TODO` to homepage, summary, and description

### DIFF
--- a/middleman-cli/lib/middleman-cli/extension.rb
+++ b/middleman-cli/lib/middleman-cli/extension.rb
@@ -25,9 +25,14 @@ module Middleman::Cli
 
     # The extension task
     def extension
+      config = {
+        git_username: git_username,
+        git_email: git_email
+      }
+
       copy_file 'extension/gitignore', File.join(name, '.gitignore') unless options[:'skip-git']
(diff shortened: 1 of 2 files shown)


tdreyno:
This is very nice. Is there a way to check if the shelling out to `git` is safe on Windows?

unasuke:
@tdreyno 

> Is there a way to check if the shelling out to `git` is safe on Windows?

In bundler, they checking `git.exe`

https://github.com/bundler/bundler/blob/fe5d24b5dc6614c0edfee8407acba14aa610ac72/lib/bundler.rb#L529-L532

Maybe Windows (cmd.exe) can use `which` too.
https://stackoverflow.com/questions/304319/is-there-an-equivalent-of-which-on-the-windows-command-line

I'll confirm later! :smiley:

PR #2303, opened 2019-12-17
URI.decode/encode is obsoleted. Use WEBrick::HTTPUtils.escape/unescape

unasuke:
URI.escape/unescape is obsoleted.
This methods are following RFC 2396.
However, RFC 2396 was obsoleted by RFC 3986.

* https://tools.ietf.org/html/rfc2396
* https://tools.ietf.org/html/rfc3986

Ruby core team says "Instead, use CGI.escape, URI.encode_www_form or URI.encode_www_form_component" but those methods are incompatible between URI.escape/unescape.

So, I choose WEBrick::HTTPUtils.escape/unescape instead, its compatible with URI.escape/unescape.

We need take off this obsoleted rfc dependent behavior in the future.

It shows many warning messages in Ruby 2.7.

### DIFF
--- a/middleman-core/lib/middleman-core/rack.rb
+++ b/middleman-core/lib/middleman-core/rack.rb
@@ -8,6 +8,8 @@ require 'middleman-core/util'
 require 'middleman-core/logger'
 require 'middleman-core/template_renderer'
 
+require 'webrick'
+
 # CSSPIE HTC File
 ::Rack::Mime::MIME_TYPES['.htc'] = 'text/x-component'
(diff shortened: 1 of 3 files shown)


tdreyno:
Thanks!

unasuke:
Thank you! and I'm sorry, I misunderstood.

I said 
> It shows many warning messages in Ruby 2.7.

, but this warning messages from Ruby 1.9. https://github.com/ruby/ruby/commit/238b979f

PR #2315, opened 2020-02-13
Remove Backports

### LINKED ISSUE #2314: Explicitly import needed features from backports
I expected Middleman to not import all backports from the `backports` gem, since this is not recommended as documented here: https://github.com/marcandre/backports#up-to-a-specific-ruby-version

Unfortunately it does, which leads to a problem with Capybara and Ruby 2.7.0. We currently run into this problem here: https://github.com/zweitag/middleman-images/pull/56

The underlying problem seems to be in `backports` and I already submitted an issue in their bug tracker, but maybe this could also be avoided by requiring only the backports, Middleman really needs.

pmk1c:
## Motivation

This resolves #2314 by removing Backports altogether.

## Changes
- Removed dependency on Backports.

## Notes

I ran all tests with Ruby 2.5.0 and they passed, so I would think Backports might not be needed anymore after all. If you know of places where it might be needed, we should add tests for that and import the needed backport explicitly.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -10,7 +10,6 @@ PATH
     middleman-core (5.0.0.rc.1)
       activesupport (~> 5.2)
       addressable (~> 2.3)
-      backports (~> 3.11)
       bundler
       contracts (~> 0.16.0)
       dotenv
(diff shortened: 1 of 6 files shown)


tdreyno:
Nice! We live in the future now!

PR #2320, opened 2020-05-03
Remove automatic alt text feature

### LINKED ISSUE #2319: Automatically generated alt text is problematic for screen reader users
## Background

Back in 2013, the Middleman community had the best intentions when building the [automatic additon of the alt-attribute](https://github.com/middleman/middleman/issues/782), inspired by the Rails community.

Automatically generating alt text based on file names is not a good idea because:

* file names are not good alt text
* this is a feature users do not expect

A digital accessibility specialist also [pointed out similar concerns to the Rails community](https://github.com/rails/rails/issues/30096) who then removed the feature.

### File names are not good alt text

Automatically generated alt text based on the file name is problematic for screen reader users. This is because the file name is never a contextualised description of the image it contains. For example, a file called `infrastructure-overview.svg` says nothing about the infrastructure itself.

### Users do not expect automatic addition of alt text

A lot of Middleman users are unaware of this feature that automatically generates alt text for them. For example, my team needed to add [a decorative image which should have an empty alt tag](https://www.w3.org/WAI/tutorials/images/decision-tree/). We thought that adding an image using `` would produce an empty alt tag. It wasn't until much later when we looked at the HTML for an unrelated reason that we noticed the automatically generated alt tag.

Adding an image to a Middleman site using both `` and `<%= image_tag "logo.svg" %>` produces:

```html
<img src="/images/logo.svg" alt="Logo">
```

This is not the expected outcome for users who are not aware of the automatically generated alt text.

## Additional information

- Ruby version: 2.6.5
- Middleman version: 4.3.6
- OS version: MacOS 10.12.6

bravegrape:
Fixes #2319


tomrutgers:
I’d rather have the option to enable automatically (opt-in) instead of deleting the functionality altogether.

bravegrape:
What is the use case you have in mind for keeping it?

tdreyno:
This would be a breaking change, so I agree with @tomrutgers. I think the safest route is included, but "off by default" for the stable releases. I'm okay removing it for the master/5.x branch

bravegrape:
@tomrutgers and @tdreyno thank you for your input on this PR 😁 

Agree that suddenly removing the feature will probably break things for some users so phasing out is the way to go 👍 

This is my first Middleman PR, so could you help me understand the next steps are?

tdreyno:
@bravegrape `master` is the 5.x line, so merging this will atleast remove it going forwards.

To phase it out on the stable branch, I'll have to do some docs and tweeting

bravegrape:
@tdreyno sounds good - I'm happy for this to be merged into master for 5.x 👍 
Should I write more information into the PR description or is the reference to the issue enough?

I'm guessing the current stable branch is `4.x` right? Shall I raise a separate PR on there to make the automatic alt text feature optional and off by default? 👀 

I'm also up for helping out with docs for phasing out, if needed 👍

tdreyno:
@bravegrape If you can make a PR to `4.x` turning the feature off by default (and keeping tests for on/off), that'd be fantastic.

PR #2322, opened 2020-05-11
Remove reference to unused reload_paths option

pwim:
The option was removed in bedf235ff68998c0fd87f7531c860f82381312c2. However, it wasn't deleted from the params hash, so a nil value is always being passed in.


tdreyno:
Appreciate it

PR #2328, opened 2020-06-02
Test building of redirects

pwim:
This is currently broken in master. Before looking further into it myself, I wonder if anyone else who is more familiar with the codebase can take a quick look, and see if it is something obvious?

Regarding my implementation of the test, I noticed that while the redirect feature makes use of the step

```
And a file named "config.rb" with:
"""
redirect "hello.html", to: "world.html"
"""
```

It seems impossible to make use of this when building an an app, so I created a new fixture. Given that I have this fixture, does it make sense to convert the rest of the scenarios to use it (i.e., is a using fixtures preferred over using the  `a file named "config.rb" with:` step)?

### DIFF
--- a/middleman-core/features/redirects.feature
+++ b/middleman-core/features/redirects.feature
@@ -60,3 +60,7 @@ Feature: Meta redirects
     And the Server is running at "large-build-app"
     When I go to "/hello.html"
     Then I should see 'hello.html to world.html'
+
+  Scenario: Build a redirect
+    Given a successfully built app at "redirect-app"
+    Then the file "build/external.html" should contain '<meta http-equiv=refresh content="0; url=http://example.com"'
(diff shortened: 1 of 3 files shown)


pwim:
I managed to fix this issue. I think it occurred any time that a `Middleman::Sitemap::Resource` was inherited and had a custom `render` method defined. `Builder` assumed any resource that was not `binary?` or `static_file?` had a `file_descriptor`. However, this was not the case for a Redirect, nor and EndpointResource.

To be honest, I didn't totally understand the intention of the FileVertex, but it seemed like the vertex tracking might not apply in this case, and so my PR bypasses it when there is no `file_descriptor`. All the tests pass, so I guess it isn't introducing any new bugs, but perhaps there's a more elegant way of handling it.

tdreyno:
Excellent. Thank you for digging in. The vertex stuff is part of the incremental build system. It is admittedly quite obtuse.

tdreyno:
The question here is whether a change to a file on disc could possibly require regenerating the redirect file. I don't think so. Only a data file or config change would do that and those are handled elsewhere.

PR #2340, opened 2020-06-11
Trigger a reload for the preview server when data changes

### LINKED ISSUE #2329: Reloading of data files doesn't work for dynamic pages in Middleman 5.0 (master)

I'm running the preview server for a Middleman app that defines proxy pages from data files. When I change the data files, these pages aren't updated unless I restart the preview server.

Suppose I have the following in `data/people.yaml`

```yaml
- name: Tom
  bio: A jolly fellow
```
and the following in `config.rb`

``` ruby
data.people.each do |person|
  proxy "/#{person.name.parameterize}.html", "/person.html", locals: { person: person }
end
```

and the following in `person.html.erb`

```erb
<h1><%= person.name %></h1>
<%= person.bio %>
```

When I start the preview server, and visit `http://localhost:4567/tom.html`, I see the dynamic page. If I change the value of `bio` in `data/people.yaml`, and reload the page, there isn't any change; however, in Middleman 4, the newest value of `bio` would be shown.

This issue appears to be limited to proxy pages. If I have an template where I iterate through all data, for example the following, reloading works properly.

```erb
<% data.people.each do |person| %><%= person.name %><% end %>
```

## Additional information

- Ruby version: 2.7.1p83
- Middleman version: 5.0.0 via master branch
- OS version: macOS 10.15.4

pwim:
Resolves #2329. Without this, proxy pages that were passed data as a local weren't being loaded when the data changed.

### DIFF
--- a/middleman-cli/features/preview_server.feature
+++ b/middleman-cli/features/preview_server.feature
@@ -553,3 +553,26 @@ Feature: Run the preview server
       """
       The Middleman is reloading
       """
+
+  Scenario: Server reloads when data changes
+    When I run `middleman server --verbose` interactively
+    And a file named "data/people.yml" with:
+      """
+      - Tom
+      """
+    And I wait for the output to contain:
+      """
+      The Middleman is running
+      """
+    When a file named "data/people.yml" with:
+      """
+      - Bob
+      """
+    And I stop middleman if the output contains:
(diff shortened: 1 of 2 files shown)


tdreyno:
Awesome. Thank you

PR #2342, opened 2020-06-15
Cleanup of how fixtures customize a fixture 

pwim:
I noticed that sometimes the same fixture was being referenced twice in a given scenario. Something like the following.

```
Given a fixture app "empty-app"
# additional customization
And the Server is running at "empty-app"
```

This is equivalent to

```
Given a fixture app "empty-app"
# additional customization
Given a fixture app "empty-app"
And the Server is running
```

So it can just be replaced with

```
Given a fixture app "empty-app"
# additional customization
And the Server is running
```

This avoids pointless duplication of the fixture name.


tdreyno:
Great. I think these are some leftover copy pasta. Thanks

PR #2348, opened 2020-08-09
Update kramdown to avoid CVE-2020-14001 in v4

le0pard:
Proposal to update kramdown dependency for security reason in v4 middleman: https://github.com/advisories/GHSA-mqm2-cgpr-p4m6


le0pard:
Thanks @tdreyno Can you release new version for v4 branch? so in this case I will point on rubygems instead git.

tdreyno:
Yes, but may be a day or two

PR #2354, opened 2020-08-16
[Proposal] Update rubocop + use enable-frozen-string-literal for project codebase

le0pard:
It is proposal to work with "enable-frozen-string-literal", which will be default behaviour for Ruby 3.


tdreyno:
Wow. This is amazing. I've been meaning to do this for a while, but could never quite get it to pass all the checks.

PR #2410, opened 2020-11-30
Add an option to modify asset_path in AssetHost extension

lsirivong:
👋 Hello!

This adds an option for a proc called `asset_path_rewriter` to the extension. 

I was hoping to add this option for the AssetHost extension because our CDN can rewrite file names to remove special characters.


tdreyno:
Seems reasonable. I'll probably merge this manually so the implementation matches how the asset_hash extension uses its proc

tdreyno:
@stale Let's keep this open

PR #2417, opened 2020-12-18
Reduce dependabot frequency

sandstrom:
Avoid too much noise in commit list, hides relevant commits in a heap of (mostly) irrelevant ones.


tdreyno:
😮 Game changing

PR #2423, opened 2020-12-27
Ruby 3.0 support

pwim:
Add dependency on WEBrick in Ruby 3.0 and above as it has been removed from the stdlib.

### DIFF
--- a/middleman-core/middleman-core.gemspec
+++ b/middleman-core/middleman-core.gemspec
@@ -27,6 +27,7 @@ Gem::Specification.new do |s|
   s.add_dependency('dotenv')
   s.add_dependency('rgl', ['~> 0.5.3'])
   s.add_dependency('toml')
+  s.add_dependency('webrick')
 
   # Helpers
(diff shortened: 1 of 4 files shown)


middleman-core/middleman-core.gemspec
```diff
@@ -27,6 +27,7 @@ Gem::Specification.new do |s|
   s.add_dependency('dotenv')
   s.add_dependency('rgl', ['~> 0.5.3'])
   s.add_dependency('toml')
+  s.add_dependency('webrick') if RUBY_VERSION >= '3.0'
```
tdreyno, line 30: I'm not sure this actually will work. I think its fine to include the dep for all versions

pwim: I did test it using Bundler's path option, but you're right that it may behave differently with a released gem, and it looks like it may be causing the error with TravisCI too. 

I also don't see any strong reason for not including the dependency everywhere, so I've amended and rebased to always require it.

PR #2425, opened 2020-12-28
Run tests using GitHub workflows

pwim:
travis-ci.org now displays "Please be aware travis-ci.org will be shutting down in several weeks, with all accounts migrating to travis-ci.com. Please stay tuned here for more information."

As GitHub actions is free for public repositories, whereas it's unclear to what travis-ci.com will be, and the switching cost is low, it seems going with workflows is the prudent thing to do.


tdreyno:
Weird. It doesn't  seem to want to run. Looks identical to working actions I have elsewhere.

pwim:
Weird, it was working for me: https://github.com/pwim/middleman/actions/runs/447987625 (though that had it run `on: push`).

tdreyno:
That worked: https://github.com/middleman/middleman/actions/runs/451172395

Issue #2433, opened 2021-01-15
Migration from MM4 to MM5 for fully i18nized site


svoop:
@tdreyno This is a followup to the mail exchange we had some days ago.

I'm trying to migrate some sites which have all their pages translated to various languages. They don't have a `localizable` directory (as suggested by [the guide](https://middlemanapp.com/advanced/localization/)), but use the source directory itself in the activation in `config.rb`:

```ruby
activate :i18n, templates_dir: '.', locales: %i(en de)
```

This works fine on MM4 but breaks on MM5. To illustrate, I've stripped one site down to the bare scaffold, prepped the `Gemfile` for both middlemen and then built with both to `build-mm4` and `build-mm5` and put it all [in this debugging repo](https://github.com/svoop/mm4to5). The result of the two builds is quite different.

[build-mm4](https://github.com/svoop/mm4to5/tree/main/build-mm4):
```
/index.html
/contact/index.html

/de/index.html
/de/contact/index.html
```

[build-mm5](https://github.com/svoop/mm4to5/tree/main/build-mm5):
```
/index.en.html
/index.de.html

/contact/index.de.html
/contact/index.en.html
```

Is this a bug or am I doing something wrong here? Thanks a bunch for your help!

tdreyno:
Does placing them inside the localizable folder fix this? I would say that the code does not expect that folder to be the root (and therefore the parent of other folders)

svoop:
Indeed, this fixes the issue.

Feels weird though having to put the entire site into yet another nesting.

Why is the `template_dir` necessary at all? From my non-dev POV, an easier way to use `i18n`:

```
# The primary locale is built into the root dir:
# activate :i18n, locales: %i(en de)

source/index.en.html
source/foo/index.en.html

build/index.html
build/foo/index.html
```

```
# Other locales are built into their path-namespace:
# activate :i18n, locales: %i(en de)

source/index.de.html
source/foo/index.de.html

build/de/index.html
build/de/foo/index.html
```

```
# Templates which are not scoped to a locale are used for every locale unless a scoped one is overriding this:
# activate :i18n, locales: %i(en de fr)

source/index.html
source/index.fr.html

build/index.html      # from source/index.html
build/de/index.html   # from source/index.html
build/fr/index.html   # from source/index.fr.html
```

I'll poke around a little in `i18n.rb`, but maybe you can give me a hint if or why I bark up the wrong tree. :smile:

tdreyno:
[...]

PR #2434, opened 2021-01-19
Support sites which are using file extension i18n only

svoop:
This implements part of what I suggested in #2433. It is best explained by what has to be rephrased at the [very end of the i18n guide](https://middlemanapp.com/advanced/localization/#localizing-entire-templates):

> ~To use this localization method, be sure to put your files inside the localizable folder.~

> To use this localization method **as well**, be sure to put your files inside the localizable folder.
>
> To use this localization method **exclusively**, disable the localizable folder with `activate :i18n, templates_dir: false`.

(The also suggested fallback (e.g. `index.html` if no `index.xx.html` is set for locale `xx`) seems to be a very rare use case and not worth the ton of code and potential for breaking existing sites.)


svoop:
Thanks, @tdreyno !

PR #2448, opened 2021-02-20
Update middleman.gemspec to rely on webrick explicitly

fiddlemath:
To solve what's probably the same issue as here: https://github.com/jekyll/jekyll/issues/8523

I've tried roughly this, locally and manually, in order to get middleman to build on ruby 3.0.0.

(I am not really a rubyist. Someone more knowledgeable should definitely double-check this before accepting it.)

### DIFF
--- a/middleman/middleman.gemspec
+++ b/middleman/middleman.gemspec
@@ -23,4 +23,5 @@ Gem::Specification.new do |s|
   s.add_dependency('haml', ['>= 4.0.5'])
   s.add_dependency('coffee-script', ['~> 2.2'])
   s.add_dependency('kramdown', ['~> 2.3'])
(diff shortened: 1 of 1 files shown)


tdreyno:
Great. thanks. Should this be backported to the stable branch as well?

le0pard:
@tdreyno I am using middleman with ruby 3.0.0 from master branch and no issue with this, because it depend from middleman-core, which already contain webrick as dependency.

@fiddlemath are you sure, that master branch has this issue?

fiddlemath:
To clarify, I did _not_ do a careful check to work out _where_ this
dependency should be inserted -- either by package or branch -- only that
it made the overall package work for me that when it was otherwise broken,
and it seemed like the right dependency to add.

PR #2324, opened 2020-05-19
Unpin ActiveSupport

brettmortensen:
Unpins ActiveSupport for compatibility with Rails 6.

PR #2462, opened 2021-04-09
Allow users to use Active Support 6.0

tnir:
We can only use activesupport 6.0.x since activesupport 6.1.0+ requires `i18n >= 1.6, < 2`: https://rubygems.org/gems/activesupport/versions/6.1.0


tnir:
cf. #2324 for middleman 5.0-rc

Issue #2482, opened 2021-06-20
Proposal: Use Rack::Handler for PreviewServer


le0pard:
Hello.

I have idea, but before contribution to it, I need confirm it is good one. 

As I can see PreviewServer using webrick (which is logical - no need additional dependency in project). But we can allow end customer to select, which webserver they want to use for preview. As I can see sinatra gem is using `Rack::Handler` to select best available web server.

I did check, and looks like middleman can work in similar way:

```diff
-        @webrick ||= setup_webrick(@options[:debug] || false)

         rack_app = ::Middleman::Rack.new(@app).to_app
-        @webrick.mount '/', ::Rack::Handler::WEBrick, rack_app
+        @server = ::Rack::Handler.get(server_information.server) || ::Rack::Handler.default
+        @server.run rack_app
```

we can provide ability for user set server from `server_information`, like `[:server] = :puma` or `::Rack::Handler.default` will try to use best one available - https://github.com/rack/rack/blob/master/lib/rack/handler.rb#L50-L59

Looks like also need more work with server configuration (ssl, logging, etc), that is why I ask if this is good idea before start contributing to it.

Thanks

tdreyno:
Seems interesting. I like the inverted control of the server used coming from Rack

le0pard:
Thanks @tdreyno I will try make as good as possible PR for this change

PR #2484, opened 2021-06-24
Use rack handler for preview server

le0pard:
This is first try to implement my proposal https://github.com/middleman/middleman/issues/2482

 - not all `::Rack::Handler` (example puma) have defined method `shutdown` and just wait for `Interrupt` error: https://github.com/puma/puma/blob/master/lib/rack/handler/puma.rb#L73 This lead, that best possible solution to run `::Rack::Handler` in fork and use `Process.kill('QUIT', server_pid)` to stop it.
 - Instead `app.logger` is used separate `logger` object, which we pass to WEBrick Logger
 - I did not check all possible options, like `ssl` and `daemon` - will try to make this later

But please leave comments or even add changes, if needed @tdreyno 

To test this simple - by default system will use `webrick`. If you will add `puma` or `falcon` in Gemfile - it will switch to this webservers. Also you can set option `server` and specify preview web server without search closest possible (default search works by array `["puma", "falcon", "webrick"]`). This option will allow to set server, which will have `Rack::Handle` class defined and available for middleman.


le0pard:
Live reload is broken - continue work

le0pard:
@tdreyno looks like my idea cannot work right now (maybe it can be done, but need more work to rewrite system, which monitor files changes). Only rack handler for webrick have method `shutdown`, but puma

https://github.com/puma/puma/blob/73c5bf28402359e5f56973977b3443ca1f3d83ae/lib/rack/handler/puma.rb#L73-L76

and falcon

https://github.com/socketry/falcon/blob/676475da61f2ad8534902c73208d5683e4180496/lib/rack/handler/falcon.rb#L38-L40

expecting web server shutdown from `exit` command. I try to make this work with `fork` usage, but in this case I continuously fighting to make work https://github.com/guard/listen gem.

Maybe later I will try again, but right now I did not found good solution without rewriting also all `listen` system in middleman.

tdreyno:
Appreciate the effort, @le0pard

PR #2487, opened 2021-07-09
Fix preview server

maxmeyer:
I tried to use the preview server https. This fails due to an SSL error.

~~~
[2021-07-09 15:48:52] ERROR OpenSSL::SSL::SSLError:
SSL_CTX_use_certificate: ca md too
weak\n\t/usr/local/bundle/gems/webrick-1.7.0/lib/webrick/server.rb:262:in
`initialize'
~~~

Along my way I also fixed a variable which is not set (anymore).


maxmeyer:
If this PR is considered to be merge,  #2484 might need some changes.

tdreyno:
Interesting. Thank you

PR #2490, opened 2021-07-24
fix middleman/ruby process 100% CPU usage problem ( in macOS 11.x Big Sur )

guanting112:
## before

before

## after

after


tdreyno:
@guanting112 did you discover a solution? we're happy to bump this dependency if it improves things?

PR #2497, opened 2021-08-14
Preview server refactoring

le0pard:
- Extracted webrick server logic in separate class
 - Use threads and queues signals instead loop for web server reload and stop logic


le0pard:
Thanks @tdreyno

PR #2517, opened 2021-12-27
Update middleman-core gemspec deps (allow activesupport 7 and ruby 3.1 usage)

le0pard:
Support ruby 3.1 - fixed Yaml safe_load. Ruby 3.1.0 installs Psych 4.0.3 as standard library and [Use Psych.safe_load by default](https://github.com/ruby/psych/pull/487)


le0pard:
I checked - everything works with ruby 3.1.0 with middleman app, but test is failing. 

Will dig later what is the issue. If you have any idea, @tdreyno about issue - write it here. Thanks

le0pard:
Ok, looks like issue with gem "contracts". Just update from 0.16.0 to 0.16.1 break tests. 

Based on docs - need to use "0.16" for Ruby 2.x and "0.17" for Ruby 3. Problem, that `RUBY_VERSION` in gemspec will not help to set dynamic gem version - https://github.com/rubocop/rubocop/issues/7137

tdreyno:
Thanks @le0pard. Given that this is the master branch and is for unreleased version 5 code, it might make sense to fully move to Ruby 3 at the same time. Let me take a few days and think about it.

middleman-core/middleman-core.gemspec
```diff
@@ -25,39 +25,39 @@ Gem::Specification.new do |s|
...
+  s.add_dependency('rgl', ['~> 0.5'])
   s.add_dependency('toml')
   s.add_dependency('webrick')
 
   # Helpers
   s.add_dependency('activesupport', ['>= 5.0.0'])
-  s.add_dependency('padrino-helpers', ['~> 0.14.4'])
-  s.add_dependency('addressable', ['~> 2.3'])
-  s.add_dependency('memoist', ['~> 0.14'])
+  s.add_dependency('padrino-helpers', ['~> 0'])
```
tdreyno, line 34: Might need to keep this more explicit. Padrino has a habit of breaking changes in their minor version... they've been at 0.x for 12 years :)

le0pard: Fixed by https://github.com/middleman/middleman/pull/2517/commits/be5871e2075f50a007b2fc6ef0846139952716cf

le0pard:
Thanks @tdreyno

Here small results:

### Changes 1

I commented some contract https://github.com/middleman/middleman/pull/2517/files#diff-5e5dbdbbafcce0e9c2e505d9ba421849c027b65571b8417657df64fa20880a61R55

Looks like it did not like `LazyCollectorStep`, which can return anything

```
can't convert Middleman::CoreExtensions::Collections::LazyCollectorStep to String (Middleman::CoreExtensions::Collections::LazyCollectorStep#to_str gives Middleman::CoreExtensions::Collections::LazyCollectorStep) (TypeError)
 /tmp/aruba/collections-app/config.rb:1:in `evaluate_configuration!'
./lib/middleman-core/application.rb:328:in `instance_eval'
./lib/middleman-core/application.rb:328:in `evaluate_configuration!'
./lib/middleman-core/application.rb:285:in `initialize'
```

I did try to fix this by adding needed methods for types:

```ruby
def to_int
  Integer(value)
end

def to_ary
  Array(value)
end

def to_hash
  Hash(value)
end

def to_str
  String(value)
end
```

but no luck:

```
undefined method `*' for nil:NilClass (NoMethodError)
      ./lib/middleman-core/core_extensions/collections/lazy_step.rb:39:in `value'
      ./lib/middleman-core/core_extensions/collections/lazy_step.rb:25:in `value'
      ./lib/middleman-core/core_extensions/collections/lazy_step.rb:51:in `to_hash'
      ./lib/middleman-core/core_extensions/collections/lazy_step.rb:39:in `gsub'
      ./lib/middleman-core/core_extensions/collections/lazy_step.rb:39:in `value'
      ./lib/middleman-core/core_extensions/collections/lazy_step.rb:25:in `value'
      ./lib/middleman-core/core_extensions/collections/lazy_step.rb:55:in `to_str'
      ./lib/middleman-core/core_extensions/collections.rb:62:in `sitemap_collector'
```

Maybe I am not fully understand logic behind `collections`, will be good, if you look.

### Changes 2

Changed contracts for `render` - https://github.com/middleman/middleman/pull/2517/files#diff-808501652dff780e70a2af6fe1e1fb6f1c8f929c4734ddc77dc8166e54188ea9R207

Maybe in some cases it called with `nil`s instead hashes, I did not spend big amount of time to check where this happening.

### Changes 3

Temporary disable 3.1 on CI - it did not want to compile `mini_racer`. I hope `mini_racer` will fix this.

### Changes 4

Now all gems require ruby >= 3.

tdreyno:
Looks good. Seems like the last remaining issue is fixing the lazy contracts... which might not be worthwhile since those are very meta.

le0pard:
Ok, I did some checks, and looks like `contracts` was not correct in some cases: `render` get args `{}`, which is hash, but `contracts` said it is `nil` and fail.

I better slowly start moving in separate PRs to RBS on Sorbet (last one have [experimental compiler](https://sorbet.org/blog/2021/07/30/open-sourcing-sorbet-compiler)), which can be useful for middleman speed (of cource, this need to be checked)

So it is now up to you @tdreyno

tdreyno:
Switching from Contracts to Sorbet would be great for 5.x

PR #2523, opened 2022-01-12
Make sass, less and coffee-script as optional renderers

le0pard:
This one need be review and merge first - https://github.com/middleman/middleman/pull/2517

For future we can extract them as `middleman-sass`, `middleman-less` and `middleman-coffee-script`, so gems will have needed deps + render interface


le0pard:
This need more work, because middleman want to compile sass for css extension too

PR #2531, opened 2022-01-30
[branch 4.x] Minify JS: migrate from Uglifier to Terser gem to support ES6

markets:
ℹ️ **The branch targets v4 branch**

Ref: #2530

Issue #2530, opened 2022-01-30
Migrate from Uglifier to Terser


markets:
ℹ️ _This is more a feature request than an issue_

---

Hi @tdreyno and Middleman community 👋🏼 

First of all thank you so much for maintaining such an incredible project, still the best SSG IMO 🤘🏼

Even the External Pipeline is awesome and I love it (just created a cool template: https://github.com/ralixjs/middleman-ralix-tailwind) for new projects, I still have other MM sites with the following setup:
 
**MM v4 + Sprockets + Bootstrap v5 gem**

Which is quite easy to get it up, without the need of a js bundler. But that setup is actually crashing when using the `:minify_javascript` setting, as Uglifier doesn't support ES6 syntax (coming from Bootstrap gem):

```
error  build/javascripts/application-ec337532.js
SyntaxError: Unexpected token: expand (...)
JS_Parse_Error.get ((execjs):3538:621)
(execjs):4077:53
(execjs):1:40
Object.<anonymous> ((execjs):1:58)
Module._compile (node:internal/modules/cjs/loader:1101:14)
Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
Module.load (node:internal/modules/cjs/loader:981:32)
Function.Module._load (node:internal/modules/cjs/loader:822:12)
Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
node:internal/main/run_main_module:17:47
/Users/markets/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/execjs-2.8.1/lib/execjs/external_runtime.rb:39:in `exec'
...
```

To fix it, I made the following changes:

```diff
diff --git a/Gemfile b/Gemfile
index afc9a85..15a2529 100755
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ gem "middleman", "~> 4.4"
 gem "middleman-livereload" , "~> 3.4"
 gem "middleman-sprockets", "~> 4.0.0"
 gem "bootstrap", "~> 5.1.0"
+gem "terser", "~> 1.1"
 
 group :development do
   gem "byebug"
diff --git a/Gemfile.lock b/Gemfile.lock
index c23ce8a..0dde00c 100755
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -163,6 +163,8 @@ GEM
       activesupport (>= 5.2)
       sprockets (>= 3.0.0)
     temple (0.8.2)
[...]

PR #2532, opened 2022-01-30
Minify JS: migrate from Uglifier to Terser gem to support ES6

markets:
Ref: #2530

Port to **v4 branch** in 👉🏼  #2531


tdreyno:
rebase and bundle update

Issue #2524, opened 2022-01-12
Release v5.0.0.rc.2?


cllns:
`v5.0.0.rc.1` was released in May 2019. 

While trying to use it, I get an error about `URI.decode`, which was fixed in #2303 in December 2019. That hasn't been released as a gem yet. 

Could we get a `v5.0.0.rc.2` release please? In the meantime, I can use the `github:` source in my `Gemfile`.

Thanks!

(Related: #2400)

Issue #2400, opened 2020-11-23
Version 5 abandoned?


cmer:
I see on that v5-rc1 was released over a year ago, but since, there was only development on the v4 line with semi-frequent releases.

Is v5 abandoned? What’s its status and should I use v4 or v5 for a brand new project?

tdreyno:
Work has been continuing on the main branch, but no gems have been published. Some folks are pointing directly at github rather than rubygems and using that version.

I wouldn't say its abandoned, there is simply not enough time/attention to coordinate, document and release a major version at this time.

Because its not out yet, v5 will likely be incompatible with many third party extensions. If you don't need those, you should be able to start with v5, then bail out back to v4 if you aren't feeling it. Going the other way, from v4 with lots of extensions would likely be more difficult right now.

cmer:
Thanks for the response.

Can we expect most extensions to not be compatible with v5, or just a few? Not sure what the changes are like under the hood.

tdreyno:
The only real change to the core that would effect extensions is the removal of using low-level Rack middleware to also page output. That API isn't really documented or advertised, so I doubt it will effect many popular extensions. Anything under the middleman org (autoprefixer, blog, syntax) will work fine

That said, the most likely difficulty will just be conflicting dependency versions. If the extension says "middleman 4" and not "middleman 4+" then the author will have to republish with a wider version range.

smoofles:
Could this info be added to the README? Might get more people to try v5, or at least understand where its at.

PR #2538, opened 2022-02-18
Revert "[branch 4.x] Minify JS: migrate from Uglifier to Terser gem to support ES6"

tdreyno:
Reverts middleman/middleman#2531

Deciding to revert for 4.x as the impact on existing projects could be very large. Will instead modify the docs to explain how to enable Terser AND set the default new project template to setup projects with Terser enabled.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -28,9 +28,9 @@ PATH
       rack (>= 1.4.5, < 3)
       sassc (~> 2.0)
       servolux
-      terser (~> 1.1)
       tilt (~> 2.0.9)
       toml
+      uglifier (~> 3.0)
       webrick
 
 GEM
@@ -207,8 +207,6 @@ GEM
     temple (0.8.2)
     term-ansicolor (1.7.1)
       tins (~> 1.0)
-    terser (1.1.8)
-      execjs (>= 0.3.0, < 3)
     thor (0.19.4)
     tilt (2.0.10)
     timers (4.0.4)
@@ -219,6 +217,8 @@ GEM
       parslet (>= 1.8.0, < 3.0.0)
     tzinfo (2.0.4)
       concurrent-ruby (~> 1.0)
+    uglifier (3.2.0)
+      execjs (>= 0.3.0, < 3)
     unicode-display_width (2.0.0)
     webrick (1.7.0)
     xpath (2.1.0)
--- a/middleman-core/features/minify_javascript.feature
+++ b/middleman-core/features/minify_javascript.feature
@@ -131,14 +131,14 @@ Feature: Minify Javascript
     Then I should see:
     """
     <script>
-      should(),all.be(),on={one:line};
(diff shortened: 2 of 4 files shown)


markets:
hey @tdreyno 👋🏼  

I understand your concerns and I also been thinking on that before submitting the PR to v4 branch, but just my 2 cents: the v5 version still not ready for next RC (as per: https://github.com/middleman/middleman/issues/2524#issuecomment-1026077774) and probably still far from final v5.0, so if we revert it on v4, there will be no released version compatible with ES6 code. This means that, for example using the official Bootstrap gem, the project won't compile by default (need to manually install Terser and pass the custom compressor). Even if we update the default template for new MM projects, for already existing projects this is gonna be a "problem".

Of course, you're the maintainer of this project and you have the latest word, but I think it's worth to try a new v4 release with Terser by default, to see if the change really breaks any MM site. We don't know when v5 will be released (as per https://github.com/middleman/middleman/issues/2400), so v4 seems it will be around for some time.

Thanks!

tdreyno:
All existing projects are working fine without ES6. This is only an issue for new projects and legacy projects who are trying to upgrade something large like bootstrap. The upgrade path for those 4.x projects is only 1 LoC to change the value in config, so the lift is very simple. If we broke all JS compatibility in a minor or patch release, people would be very surprised.

markets:
Not sure how the new compressor can "broke all JS compatibility", AFAIK, Terser also minifies well ES5 code. I migrated (to Terser) like 6 MM sites in the last months with no problems so far! Of course, 6 is not a relevant number in that context, but some of those sites are not so small, and everything worked very smooth.

tdreyno:
That's good to know. I had issues with Terser when it was first released, but those may be resolved now. At the very least, as shown by the PR test changes, the output does alter slightly which would still be surprising in a patch release.

It's slow, but if we change the default for new projects to Terser and its chill for many months we could do a minor version release changing the default for 4.x

markets:
> the output does alter slightly which would still be surprising in a patch release

Yep, I agree that a `patch` release is not the best option for that kind of change, but probably a `minor` is acceptable? Those changes seems caused by the new Terser default for "IIFE" design pattern (so Rails apps for example, are also running with these defaults now, which I assume it is just fine 🤷‍♂️).

> It's slow, but if we change the default for new projects to Terser and its chill for many months we could do a minor version release changing the default for 4.x

Yeah! that's a bit slower but safer. That's ok too! but since the final v5 release is quite "uncertain" at this point (as you pointed out in https://github.com/middleman/middleman/issues/2539#issuecomment-1049207377), a v4 release with ES6 enabled by default doesn't seem a bad idea.

shaunakv1:
@tdreyno  I understand why you decided to not roll it into v4.x, but as you mentioned we can still enable it if needed? Did you ever get around to updating the documentation?

Thanks Much!

PR #2541, opened 2022-03-16
use YAML.unsafe_load to support Ruby 3.1/Psych 4

gyugyu:
Since `YAML.load` has been changed to an alias of `Psych.safe_load` from Ruby 3.1/Psych 4.0, using Date, Time, etc. cause raise owing to prohibiting of class loading at frontmatter.
https://github.com/ruby/psych/pull/487
To avoid this problem, use `YAML.unsafe_load` if it is definded (`YAML.load` is renamed to `YAML.unsafe_load`).
I think this is better implementation because

- This frontmatter comes from developer's input, so it may be treated as safe input.
- This modification will not break previous behavior of Middleman.

but you can choose passing `permitted_classes` option to `YAML.load` to fix vulnerability.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     name: ubuntu-ruby-${{ matrix.ruby-version }}
     strategy:
       matrix:
-        ruby-version: [3.0, 2.7, 2.6, 2.5]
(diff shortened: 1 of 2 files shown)


markets (approved):

markets:
Hi @gyugyu 👋🏼 That seems fine! But in order to get a green ✅ build, could you please rebase `middleman:4.x`?

PR #2551, opened 2022-07-07
Replace sassc with sass-embedded

ntkme:
The `sassc` gem has been deprecated in Oct 2020. [`sass-embedded`](https://github.com/ntkme/sass-embedded-host-ruby) is a new ruby api implementation that uses the official [dart-sass-embedded](https://github.com/sass/dart-sass-embedded) compiler internally.


tnir:
👍
Once this is merged to 5.x, I hope this is backported to 4.x.

Gemfile
```diff
@@ -25,9 +25,10 @@ gem 'haml', '>= 4.0.5', require: false
 gem 'kramdown', '~> 2.3', require: false
 gem 'liquid', '>= 3.0', require: false
 gem 'redcarpet', '>= 3.1', require: false
-gem 'sassc', '~> 2', require: false
+gem 'sass-embedded', '~> 1', require: false
 gem 'sinatra', '>= 2.0.0.beta2', require: false
 gem 'slim', '>= 2.0', require: false
+gem 'tilt', github: 'rtomayko/tilt'
```
ntkme, line 31: This tilt gem override is currently needed or we have to install one of sassc/sass gem. I have asked in https://github.com/rtomayko/tilt/pull/367 to see if we can get tilt to cut a new release so that it won't panic.

ntkme:
@tdreyno This PR is ready for review. Please let me know if you have any feedbacks or questions.

ntkme:
Not sure what's the issue with failed macos ruby 3.1 CI test. The issue does not seem to be related to this change. Also, all tests worked fine for me on linux ruby 3.1 locally.

middleman-core/features/asset_hash.feature
```diff
@@ -112,7 +112,7 @@ Feature: Assets get file hashes appended to them and references to them are upda
     Given a fixture app "asset-hash-host-app"
     And a file named "config.rb" with:
       """
-      set :sass_source_maps, false
+      config[:sass][:source_map] = false
```
tdreyno, line 115: This is a user-facing API change. Can we avoid this?

middleman-core/fixtures/sass-assets-path-app/config.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-set :sass_assets_paths, [
+config[:sass][:load_paths] = [
```
tdreyno, line 3: This is a user-facing API change. Can we avoid this?

tdreyno:
5.x is probably abandoned. Commented on not breaking config api compat. Might be best to re-target at 4.x/main

middleman-core/fixtures/sass-assets-path-app/config.rb
ntkme: Done.

middleman-core/features/asset_hash.feature
ntkme: Done.

ntkme:
Note, this PR relies on `sass-embedded ~> 1.70`, which requires `ruby >= 3.1`.

markets:
@ntkme Do you think we can use a version which supports current CI matrix (Ruby 2.5+)? The idea is to keep compatibility for all those versions, so if that is not possible, we'll probably need to keep both code branches under some check.

ntkme:
In general, I'm in favor of dropping support of end-of-life ruby versions, as they become burden of adopt new features.

The compatibility policy of `sass-embedded` gem is that, at minimum it would target all MRI releases in "normal maintenance" status (see https://www.ruby-lang.org/en/downloads/branches/), plus the latest JRuby stable branch. When `sass-embedded=1.70.0` was released, all MRI versions `< 3.0` were end of life, `3.0` was in security maintenance and would be end of life in 2 months, and JRuby was at 9.4, which is MRI 3.1 compatible. Therefore, that release was targeting MRI 3.1.

Ruby 2.5 was never supported in any stable versions of sass-embedded. The very last version supporting ruby 2.6 was 1.58.3, which is missing lots of features and bug fixes. Specifically, there are also a few features added recently to make this middleman PR smaller.

---

An alternative is to still install `sassc` by default, and create a new option that allows user to opt-in with using `sass-embedded`. For reference, for Jekyll we went with this approach for [jekyll-sass-converter 2.2.0](https://github.com/jekyll/jekyll-sass-converter/blob/v2.2.0/README.md#sass-embedded), and then later dropped `sassc` completely in [jekyll-sass-converter 3.0.0](https://github.com/jekyll/jekyll-sass-converter/blob/v3.0.0/README.md#dropped-implmentation-option). `jekyll 4.3` is locked to `"jekyll-sass-converter", ">= 2.0", "< 4.0"`, so that user can still choose their preferred implementation by choosing between jekyll-sass-converter 2.x or 3.x in `Gemfile`.

I'm not recommending this option, because it's a lot of extra work supporting end-of-life software.

tdreyno:
I would recommend having two renderers. They get enabled if the gem they depend on is present (there's other examples of this in the code where adding a templating engine not in the default gem spec will enable support). 

https://github.com/middleman/middleman/blob/main/middleman-core/lib/middleman-core/core_extensions/rendering.rb#L9-L13

If the require throws (missing gem), the renderer will not enable.

The concept of "end of life" doesn't make a lot of sense for pieces of tooling like Middleman. Folks have 15 year old blogs running just fine and they shouldn't need to worry about upgrading ruby. Old code should continue to just work.

We can update the new project template to include sass-embedded for new projects and users who want to and can switch can add it to their gem file. We can also bump the minor version as a way to treat this as a new feature and encourage upgrades

deivid-rodriguez:
> The concept of "end of life" doesn't make a lot of sense for pieces of tooling like Middleman. Folks have 15 year old blogs running just fine and they shouldn't need to worry about upgrading ruby. Old code should continue to just work.

This feels strange to me. Old blogs will keep working normally, even if the minimum supported Ruby is bumped. They will just not be able to upgrade to the version of middleman that first drops support, but that does not seem like an issue. Once can make the same argument for middleman itself: "folks having 15 years old blogs running just fine shouldn't need to worry about upgrading middleman".

deivid-rodriguez:
I actually just noticed that Ruby 2.5 and Ruby 2.6 support were [dropped](https://github.com/middleman/middleman/pull/2738) recently, so only Ruby 2.7 are Ruby 3.0 are causing issues now.

ntkme:
Ruby 2.7 and 3.0 already reached EOL, which is why their support has been dropped in sass-embedded:

https://www.ruby-lang.org/en/downloads/branches/

SassC for Ruby also reached EOL:

https://github.com/sass/sassc-ruby/blob/master/README.md#sassc-has-reached-end-of-life

deivid-rodriguez:
Yeah, I think the original policy of keeping support for old rubies has been revisited, so maybe there's a chance to drop support for 2.7 and 3.0 as well.

deivid-rodriguez:
Ruby Release policy is not really semver but a modified version of it, where 2.7 and 3.0 are essentially the same. They could be both breaking, but they are usually not in practice. See https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/.

I still don't understand why bumping the minimum supported version would make existing sites lose functionality or be forced to change their CSS output. They will not be able to upgrade middleman unless they also upgrade Ruby, but they should keep running normally if they don't upgrade either.

deivid-rodriguez:
Well, in that link I shared it's stated that major versions are reserved for backwards incompatible changes, so it's actually more similar to semver than I thought, but I'm not sure what made Ruby 3.0 more breaking than other releases. In fact, this excerpt from Release [announcement](https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/) suggests the opposite and that was my experience at the time too:

> Ruby3.0 is a milestone. The language is evolved, keeping compatibility. But it’s not the end. Ruby will keep progressing, and become even greater. Stay tuned! — Matz

ntkme:
> Existing sites should not lose functionality or be forced to change their CSS output in a non-major release.

In my opinion, technology moves on. Users always have the choice to run “old versions” and never update at all if they choose to do so, especially in the case of static sites generators where security updates is generally not much of a concern.

We can look at Jekyll and GitHub Pages as a similar case here. Jekyll has migrated from the initial sass, to sassc, and then to sass-embedded over the course of the years. However, GitHub Pages classic, for the sake of being fully compatible of “existing sites”, is still locked to Jekyll 3.7 and original ruby sass. Today, many old and new sites still run on the outdated GitHub Pages classic, with the option to use new GitHub Actions based Pages and any version Jekyll and sass.

Does middleman need a major version? Preferably yes, but I think there is a better way to handle this. For Jekyll, the sass support is split as a separate gem called jekyll-sass-converter, for which 1.x is sass, 2.x is sassc, and 3.x is sass-embedded. The main Jekyll gem, currently at 4.x, allows jekyll-sass-converter >=2.0, <4.0, effective allows end users to choose sassc or sass-embedded by simply pinning dependency version in Gemfile.

---

Jekyll and GitHub Pages today have a much larger user base and honestly we haven’t seen any major complaints about the sass-embedded upgrade, except a few common issues:

1. Bugs in old version of rubygems/bundler causing incorrect version of sass-embedded to be installed and crash. This is one of the reasons for dropping older rubies so that it’s less likely to encounter a buggy rubygems/bundler.
2. BSD is not supported yet, but will be supported soon.
3. Deprecation warnings and upcoming breaking changes. Sass language has changed over the years. Some features will have a breaking change in the upcoming sass-embedded 2.x and 3.x, but as of current 1.x, it has extremely good compatibility with existing syntax, that nearly no changes are required as of today. All the noisy warnings can be silenced and users have choice to lock themselves at old version if they like.

The take away from jekyll-sass-converter upgrade as a contributor is that: so far we have seen lots of complains about depreciation warnings being too noisy, but as they are just warnings and the actual behavior has not changed, the sass syntax compatibility is not really an issue for now. - However, keep in mind that once users update to the new syntax, there will be no more going back, because sassc has not been maintained and there is absolutely no forward compatibility with new features in sass-embedded. The lack of forward compatibility can potentially be a bigger trouble than backward compatibility for non-technical users who cannot figure out that they are running different versions on local and on server. E.g. running Jekyll 4 with sass-embedded locally using new syntax and site might be broken on GitHub Pages classic, which runs Jekyll 3 with sass.

In other words, in reality we have more users encounter issues with old version not compatible with new syntax, instead of new version not compatible with old syntax, which does happen in some edge cases but extremely rare.

deivid-rodriguez:
> Existing sites should not lose functionality or be forced to change their CSS output in a non-major release.

Oh, after reading @ntkme's comment I get what you mean. I think it'd be perfectly fine to release this PR in a major release.

ntkme:
By the way, there is a way to use sass-embedded today with middleman, by simply adding the following lines to `Gemfile`:

``` ruby
gem 'sassc', github: 'sass/sassc-ruby', ref: 'refs/pull/233/head'
gem 'sassc-embedded'
```

See: https://github.com/sass-contrib/sassc-embedded-shim-ruby/blob/main/README.md

This allow users who are interested to test out things earlier and learn about new sass changes.

ntkme:
I decided to go with less disruptive route of simply replace `sassc` with `sassc-embedded`.

ntkme:
Not sure why live preview tests in `asset_hash.feature` is giving me a different hash than GitHub Actions... I tried on both mac and debian, and I get a consistent hash locally. I've pushed a new commit to just update the hash to the one expected in GitHub Actions.

That aside, please be aware that the CI tests will fail on Ruby <3.1 because sass-embedded requires Ruby >=3.1 (technically even ruby 3.1 is EOL).

vmsp:
Are we moving ahead with this change? I want to start a new project but it doesn't make sense to do so using a dead version of sass.

ntkme:
> Are we moving ahead with this change? I want to start a new project but it doesn't make sense to do so using a dead version of sass. Jekyll swapped sassc for sass-embedded in 2022.

I think the main problem is that middleman still want to support ruby 2.7 which has been end-of-life for years. `sass-embedded` currently only support ruby >=3.1, effectively require middleman to drop ruby 2.7 and 3.0 support if we want to move with this.

PR #2603, opened 2022-12-19
Fix url_for with i18n when "mount_at_root" is false

### LINKED ISSUE #2594: i18n url_for doesn't produce correct link when "mount_at_root: false"

`config.rb`:

```rb
activate :i18n, :langs => [:en, :es], :mount_at_root => false
```

Given this config and some localized template, e.g.: `localizable/path.html.erb`, using `url_for`/`link_to` helpers result in wrongly generated links.

Let's say we have to get the "path" page link:

```erb
<%= url_for"/path.html", locale: :en %>
```

This would generate `/path.html` in master branch instead of `/en/path.html` as it should be (this is correct for `4.4.3` version) because the path can not be translated into a proper reference so it returns `path.html`. This bug was introduced in #2263 (tested with commit previous to the merged one) with a new lookup table implementation. In this example, the table, which I printed modifying middleman source code, would be:

```json
{
  "/en/path.html": {
    "en": "/en/path.html",
    "es": "/es/path.html"
  },
  "/es/path.html": {
    "en": "/en/path.html",
    "es": "/es/path.html"
  }
}
```

Clone this [test repository](https://github.com/phvega/mnt_at_root-test) and modify `mount_at_root` option between `:en`/`:es` and `false` to see the different links generated.

## Additional information

- Ruby version: ruby 2.6.8p205 (2021-07-07 revision 67951) [x86_64-linux]
- Middleman version: `5.0.0.rc.1` (and `master`)
- OS version: Ubuntu 20.04.1

guillerodriguez:
This fixes #2594.

### DIFF
--- a/middleman-core/features/i18n_link_to.feature
+++ b/middleman-core/features/i18n_link_to.feature
@@ -267,3 +267,97 @@ Feature: i18n Paths
     And the Server is running
     When I go to "/index.html"
     Then I should see '<a href="https://www.example.org/">Example</a>'
+
+  Scenario: Using url_for with the no mount config
+    Given a fixture app "empty-app"
+    And a file named "data/pages.yml" with:
+      """
+      - hello.html
+      """
+    And a file named "locales/en.yml" with:
+      """
+      ---
+      en:
+        msg: Hello
+      """
+    And a file named "locales/es.yml" with:
+      """
+      ---
+      es:
+        paths:
+          hello: "hola"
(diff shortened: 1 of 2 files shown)


tdreyno (approved):

guillerodriguez:
@tdreyno Just pushed a minor change to the test code, could you approve running the workflows?

tdreyno (approved):

PR #2622, opened 2023-02-04
Use Addressable instead of Webrick for encode/unencode

Juanmcuello:
This is a backport of 433a4242e411e9f70077aeff5fca51f68b2d0120 from master, but removing the unrelated changes in the Gemfile.lock.

I particularly need this because the use of  `WEBrick::HTTPUtils.unescape` (introduced in 7c155c298a445f1cc2fb0154721cee260ba333a8) is not totally compatible with `URI.decode`, that was used before, because it does not support unicode chars in the URL. 

 `Addressable::URI.unencode`, already in `master`, works as expected. Ir would be great to backport that change to this branch (`4.x`).

### DIFF
--- a/middleman-core/lib/middleman-core/builder.rb
+++ b/middleman-core/lib/middleman-core/builder.rb
@@ -1,3 +1,4 @@
+require 'addressable/uri'
 require 'pathname'
 require 'fileutils'
 require 'tempfile'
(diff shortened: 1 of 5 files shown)


markets (approved):

Juanmcuello:
@markets , I added a few more changes in the other places where WEBrick was being used. Tests should pass now.

markets:
Thanks @Juanmcuello 🙌🏼! 

Unfortunately I don't have permissions to allow CI runs or merge pull requests. Let's ping @tdreyno.

Issue #2657, opened 2023-09-09
Missing traversal data


EtienneMiret:
## Expected behavior and actual behavior

According to [Middleman: The Sitemap](https://middlemanapp.com/advanced/sitemap/#accessing-the-sitemap-from-code), I should be able to access [Middleman::Sitemap::Extensions::Traversal](https://www.rubydoc.info/gems/middleman-core/Middleman/Sitemap/Extensions/Traversal) data using the `current_resource` variable.

However, while this variable is available in my template, the `parent` is usually `nil`, while `children` and `siblings` are empty arrays.

## Steps to reproduce the problem (from a clean middleman installation)

Template:

```erb
Path: <%= current_resource.path %>

Siblings: <%= current_resource.siblings %>
Children: <%= current_resource.children %>
Parent: <%= current_resource.parent %>
```

Output:
```
Path: foo/index.txt

Siblings: []
Children: []
Parent: 
```

## Additional information

- Ruby version: ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-darwin22]
- Middleman version: Middleman 4.5.0
- OS version: Mac OS Ventura Version 13.5.1 (22G90)

tdreyno:
Can you share the full project source? Can you also post expected output

EtienneMiret:
I’m not sure I want to share the full project. And since it is ~ 650MiB you probably don’t want it anyway (don’t worry, it’s mostly assets, I don’t have that muc source code!).

However I was able to reproduce the issue in a test case, see #2660 and [CI run 6165401880](https://github.com/EtienneMiret/middleman/actions/runs/6165401880).

It seems this issue only happens on non-HTML files. If that’s the case, maybe this is because of a setting I don’t know about? My test case uses plain text and log files, but in my real project I only have XHTML files. So I would like this feature to be enabled at least for this specific file type, but I see no reason why it wouldn’t be enabled for any type.

EtienneMiret:
As a side note, I worked around this by querying sitemap and filtering on resource paths.

tdreyno:
I think it's because the code is using the global "index file" to figure out the paths. 

https://github.com/middleman/middleman/blob/e1bc04e9a3f17c71c927e5a2baf8aab5a7e2de7f/middleman-core/lib/middleman-core/sitemap/extensions/traversal.rb#L25

This is set here: 

[...]

PR #2660, opened 2023-09-12
Test that traversal data is set on non-html files

EtienneMiret:
Adds some test cases for #2657.


tdreyno:
Superseded by https://github.com/middleman/middleman/pull/2661

PR #2528, opened 2022-01-15
External pipeline manifest helpers

le0pard:
## Problem

External pipeline is awesome feature, which allow to use any JS/CSS build tool together with middleman. 

Main problem, that often such tools even in dev mode can provide bundeled assets with `[hash]` in names, like `app-553d121b3246dc3ea415.css`, `app-6427c0e37380b2306421.js`. To get information how to find assets, this tools generate `manifest` json file, which contain mapping where to find bundled files. Examples:

 - Webpack can use `webpack-assets-manifest` package. Example in config:

```js
...
plugins: [
  new WebpackAssetsManifest({
    output: 'manifest.json',
    publicPath: config.output.publicPath,
    writeToDisk: true,
    integrity: true,
    integrityHashes: ['sha256']
  }),
  ...
]
...
```

 - Vite provide this as option: https://vitejs.dev/guide/backend-integration.html

```js
export default defineConfig({
  build: {
    // generate manifest.json in outDir
    manifest: true,
    rollupOptions: {
      // overwrite default .html entry
      input: '/path/to/main.js'
    }
  }
})
```

Middleman do not have mechanism to get info from manifests, so users need develop own. 

## Solution

This changes added new option `manifest_json` to external pipeline extension and created tree additional helpers: `external_pipeline_manifest_value`, `external_pipeline_javascript_tag` and `external_pipeline_stylesheet_tag`.

### How to use it

We need to define external pipeline with `manifest_json` option:

```ruby

assets_dir = File.expand_path('.assets-build', __dir__)

activate :external_pipeline,
         name: :webpack,
         command: "yarn run assets:#{build? ? 'build' : 'watch'}",
         source: assets_dir,
         latency: 2,
         ignore_exit_code: true,
         manifest_json: File.expand_path('manifest.json', assets_dir)

```

For example, in results we get manifest.json with such content:

```json
{
  "app.css": {
    "src": "/app-553d121b3246dc3ea415.css",
    "integrity": "sha256-owU0/VSxjr/sZKK0WZQJzevzgcUEzXvPcBZFiUGzJFg="
  },
  "app.js": {
    "src": "/app-6427c0e37380b2306421.js",
    "integrity": "sha256-0PLWRauwxSW9A9b9VOa9TPyDj3gRGfxyKRhTVBKN0EM="
  },
}
```

As we can see, we can get path to `app.css` by json path `["app.css"]["src"]` and path to `app.js` by json path ``["app.js"]["src"]``. We can use our new helpers:

```ruby
[...]

PR #2663, opened 2023-10-10
Support more types being extracted from YAML

h-lame:
# What

For platforms that use YAML::safe_load we have to say which types are supported otherwise the data is rejected when loaded.  In adding ruby 3.1 support (see #2635) we allowed Date and Symbol classes, however in the 5.x branch we support Date, DateTime, Time, Symbol and Regexp by default (see #2528 - specifically 74b01d768b568fd35d163390ead59ad2a09bddd2).  This commit brings the same set of types into the 4.x branch.

# Why

We use middleman for the [lrug.org](https://github.com/lrug/lrug.org) site and while exploring upgrading to ruby 3.2 everything broke because we have timestamps in the front matter.  The simplest solution (short of us rewriting all our frontmatter) seems to be extending the support already present in 4.x to allow a few more types that YAML will automatically extract if you let it.  The 5.x branch has a wider set of classes, and so we assume that's a safe enough set to use.

# Questions

I added a feature file to provide regression testing for this feature, to make sure that the set of types we support from YAML will continue working if new ruby versions change things up further.  Is that the best way to to this?  Should it be a spec instead?

### DIFF
--- a/middleman-core/features/front-matter.feature
+++ b/middleman-core/features/front-matter.feature
@@ -80,6 +80,27 @@ Feature: YAML Front Matter
     When I go to "/front-matter-change.html"
     Then I should see "Hello World"
 
+  Scenario: A template should handle YAML with types
+    Given the Server is running at "frontmatter-app"
+    And the file "source/front-matter-types.html.erb" has the contents
+      """
+      ---
+      date_matter: 2020-01-01
+      time_matter: 2020-12-31 13:14:15 +0000
+      symbol_matter: :a_symbol
+      regexp_matter: !ruby/regexp /a|b/
+      ---
+      Date: <%= current_page.data.date_matter.strftime('%d %m %Y') %>
+      Time: <%= current_page.data.time_matter.strftime('%d %m %Y %H::%M::%S %z') %>
+      Symbol: <%= current_page.data.symbol_matter.is_a? Symbol %>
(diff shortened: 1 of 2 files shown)


chadlwilson:
This would be useful to have, especially `Time` and `DateTime` which are issues for me with an older middleman site.

Any chance of some help here @tdreyno ?

h-lame:
Looks like this could be merged if it passes CI.  Thanks for enabling auto-merge  @tdreyno!  However, I think the test failures are all present on main so it won't pass.  

The ruby 2.5 or 2.6 builds won't install, and the rest look to be failing with a bunch of errors like:

      can't convert Middleman::CoreExtensions::Collections::LazyCollectorStep to String (Middleman::CoreExtensions::Collections::LazyCollectorStep#to_str gives Middleman::CoreExtensions::Collections::LazyCollectorStep) (TypeError)

Which, as far as I can tell is coming from the `contracts` gem - I think something has changed about `BasicObject` and `to_s` or `inspect` so there may or may not be some failure, but `contracts` is hiding it because it can't compare things.  I haven't looked too deeply though so could be wrong.

What can we do to help get this merged?  I'm happy to explore fixing the build on `main` but I don't want to step on any toes.

tdreyno:
Confused how versions can change without Gemfile.lock changing. But this should be unrelated. I'll double check locally

h-lame:
Thanks for merging @tdreyno !

PR #2670, opened 2024-01-16
Allow users to use newer versions of Active Support

amatsuda:
Let's see if the tests pass...


markets:
Thanks @amatsuda!

Supersedes #2664

tdreyno:
Seems fine since it was already working with 7.0

PR #2676, opened 2024-01-19
Bundle update aruba and cucumber

amatsuda:
This patch updates aruba version restriction from `~> 1.0.0` to `~> 1.0`, and unlocks cucumber for more updates to come.

Also, this update **downgrades** activesupport that were once upgraded via #2672 but doesn't work well with currently released versions of some cucumber dependencies (see: https://github.com/middleman/middleman/actions/runs/7580742081/job/20647066221 ).


markets:
Thanks @amatsuda! that makes sense to me ✅

tdreyno (approved):

PR #2678, opened 2024-01-19
Bundle up capybara, rspec, and minitest

amatsuda:
Newer is better, I guess.


markets (approved):
As long as CI passes for all supported Ruby versions and deps are only related to development/testing... Why not?

PR #2685, opened 2024-01-24
Use contracts ~> 0.17 when ruby version is at least v3

yapcm96:
`contracts` should use v0.17 (now that it is available) when using Ruby version >= 3.0.0, since if using Ruby v2 then `contracts` v0.16 should be used.

Conditionals that was seen in previous commits in the middleman-core Gemfile have been readded in order to install the appropriate `contracts` gem version based on Ruby version used.

If there is anything missing or glaring that I might not be aware of, please point it out!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,3 +46,9 @@ gem 'middleman-core', path: 'middleman-core'
 
 # gem 'middleman-compass', github: 'middleman/middleman-compass', require: false
 # gem 'middleman-sprockets', github: 'middleman/middleman-sprockets', require: false
+
+if RUBY_VERSION < '3.0.0'
+  gem 'contracts', '~> 0.13', '< 0.17'
+else
+  gem 'contracts', '~> 0.17'
+end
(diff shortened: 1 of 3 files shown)


markets:
Hey @yapcm96 👋🏼  

Sorry, not sure what problem are you trying to solve 🙏🏼, I'm already using MM with Ruby 3 and the current `contracts` version with no problems...

yapcm96:
Hello @markets 

Regarding your question:
"What problem are you trying to solve? I'm using MM with Ruby 3 with current contracts version with no problems."

In our project, when doing `bundle update` with `gem 'middleman-gh-pages', '~> 0.4.1'` in the Gemfile of our project we get the following post install message:

```
Post-install message from contracts:

    0.16.x will be the supporting Ruby 2.x and be feature frozen (only fixes will be released)
    For Ruby 3.x use 0.17.x or later (might not be released yet)     
```                                                                                                                                                                                                           

Our Gemfile.lock is showing version 0.16.1 for `contracts` and we are using ruby version 3. I think it would be good to use 0.17.x as it'll mean that we're not locked on fixes only on 0.16.x and we'll be able to use any future features that may be added to `contracts` in the future.

Otherwise we would have to specify `gem 'contracts', '~> 0.17'` in the Gemfile of our project.

PR #2686, opened 2024-01-27
Bundle update aruba and cucumber

amatsuda:
This patch updates both aruba and cucumber to the newest released versions.

Just updating the gems caused some errors, and so I needed to add the second commit which may be a little bit tricky. So let me explain.

The Cucumber `World` instance in our feature specs extends both `Capybara::DSL` and the RSpec matchers via `capybara/cucumber',
https://github.com/teamcapybara/capybara/blob/a210c6e57872416ca689091614e65163e2886694/lib/capybara/cucumber.rb#L7-L8

and in such case, calling `RSpec::Matchers#all` method inside features or step definitions (in our case, [here]( https://github.com/cucumber/aruba/blob/b9d0f15d1292697b051f94984925b4f63c4efac4/lib/aruba/cucumber/file.rb#L98)) would fail in recent versions (after https://github.com/cucumber/aruba/commit/ac8ccef9404921fa1d63f703905bf6aeb1e5c10d) because these two modules both implement `all` method, and the RSpec one is overridden by Capybara's.

To workaround this, Capybara provides a monkey-patch module for switching these two `all` methods i.e. Capybara::RSpecMatcherProxies:
https://github.com/teamcapybara/capybara/blob/a210c6e57872416ca689091614e65163e2886694/lib/capybara/rspec/matcher_proxies.rb#L5-L11
This module is configured to be mixed-in via `included` or `prepended` hooks on Capybara::DSL and RSpec::Matchers by default,
https://github.com/teamcapybara/capybara/blob/a210c6e57872416ca689091614e65163e2886694/lib/capybara/rspec/matcher_proxies.rb#L5-L11
but since Cucumber `extend`s these modules into World instance, the module was not actually mixed-in in our case.

For reference, cucumber-rails achieves this integration by violently `include`ing RSpec::Matchers into Cucumber core Class,
https://github.com/cucumber/cucumber-rails/blob/main/lib/cucumber/rails/rspec.rb
but to me that seems a little bit too rude, so instead, let us here simply cherry-pick the only module that we need into our own World instance.

Anyway, now that the CI is all green, I think this patch is ready to be merged.


tdreyno (approved):

markets:
@tdreyno let's merge this one ✅

tdreyno:
Seems like this should already be true: https://github.com/middleman/middleman-blog/blob/master/features/published.feature

markets:
hey @tdreyno I think your last comment is related to this PR: https://github.com/middleman/middleman-templates-blog/pull/2

tdreyno:
approved. needs rebase

amatsuda:
@tdreyno Rebased!

PR #2709, opened 2024-06-25
Add support for i18n > 1.6.x

pezholio:
This adds support for i18n versions greater than `1.6.x` as we have some dependency issues upstream. There's no major breaking changes that I can see, so this should be safe to update. 

I noticed that the [dependabot PR](https://github.com/middleman/middleman/pull/2703) for the latest version of i18n was breaking for Ruby 2.5, so have added a conditional to stick with `~> 1.6.0` for that older Ruby version. Arguably 2.5 is EOL, but I didn't want to make any rash (and potentially controversial!) decisions about dropping support for older Rubies


markets (approved):
Seems fine to me. Thanks @pezholio!

tdreyno:
@pezholio thanks. this is fantastic. I think that 2.5 issue was actually a CI flake. It should be fine to remove the conditional

tdreyno:
Actually, just merged the dependabot PR instead. Conditionals won't work in gemspec and I think the approach that change took of widening the range works better. Let me know if that change solves your dependency problem @pezholio

markets:
@tdreyno I think we are all ok to close this one now, since you already merged https://github.com/middleman/middleman/commit/7301bbbacc61eb8af7eb02f8135dfa8043f8594f

PR #2739, opened 2024-10-09
Gemfile cleanup: these gems were added to test old Rubies

markets:
Since now the minor Ruby version is 2.7, we can actually remove those gems added to test old Rubies.


markets:
@tdreyno another cleanup ready ✔️

tdreyno (approved):

PR #2749, opened 2024-10-30
Strip down container gem to minimal structure

markets:
The `middleman` gem is only like a container for `middleman-core` and `middleman-cli`, so I think it's a good idea to make the structure as minimal as possible. Since I started contributing to MM, I always found the current folders a bit confusing... like if something was missing there.  

**EXTRAS**
- Remove Compass leftovers
- `*.gemspec` simplifications
- Clean up `.yardopts` files


markets:
@tdreyno ready for review! The pending failure on Ruby 3.2 is an intermitent test, unrelated to these changes.

tdreyno (approved):
lgtm. thanks!

PR #2750, opened 2024-10-30
Move runtime deps to middleman-core gem

markets:
And remove them from "Optional middleman dependencies, included for tests" section, actually they are not optional, but included with MM.


markets:
@tdreyno Are you ok with this one?

PR #2751, opened 2024-10-31
[deps] Delete unnecessary "backports" runtime dep

markets:


markets:
@tdreyno ready for review 👀

Issue #2619, opened 2023-01-24
New Slim Version 5.0


lraboteau:
Hi guys! The Brand new slim version (5.0) pushed on the 23rd of january broke my pipeline. It looks like we have parsing issues with it. Downgrade to version 4.1 on your Gemfile to temporarily fix the issue.

Atelier-Mirai:
I am also experiencing the same event.

It was also discussed in Slim.
https://github.com/slim-template/slim/issues/909

It seems to be caused by the code in middleman-core that does the rendering of slim.
https://github.com/middleman/middleman/blob/3812b031b6313ade4622c01f95ca154f36e09fb7/middleman-core/lib/middleman-core/renderers/slim.rb#L42

``` ruby
def initialize(_app, _options={}, &_block)
        super

        # Setup Slim options to work with partials
        ::Slim::Engine.disable_option_validator!
        ::Slim::Engine.set_options(
          buffer: '@_out_buf',
          use_html_safe: true,
          generator: ::Temple::Generators::RailsOutputBuffer,
          disable_escape: true
        )
      end
```


It worked by commenting out the following line

``` ruby
          generator: ::Temple::Generators::RailsOutputBuffer,
```

I hope it will be fixed.

tdreyno:
> Downgrade to version 4.1 on your Gemfile to temporarily fix the issue.

Downgrade to 4.1 in your Gemfile permanently.

kuboon:
https://github.com/middleman/middleman/issues/2619#issuecomment-1447278358
this comment-out worked for me.
Can I make PR to remove this line?

tdreyno:
@kuboon as long as that change doesn't break support for Slim 4

noraj:
It would be nice to update title to at least `~> 2.1` on middleman-core, so that middleman could allow slim 5.1+ support.

https://github.com/middleman/middleman/blob/34d0160bfcc1b717660a9a604ce7c30bb0f56ec5/middleman-core/middleman-core.gemspec#L24

https://github.com/middleman/middleman/blob/34d0160bfcc1b717660a9a604ce7c30bb0f56ec5/Gemfile#L25

Tilt 2.1 [changelog](https://github.com/jeremyevans/tilt/blob/master/CHANGELOG.md) doesn't look to have breaking changes.

That could help fix incompatibility with modern slim versions.

```
➜ bundle install                                                       
Fetching gem metadata from https://rubygems.org/............
Resolving dependencies...
Could not find compatible versions

[...]

PR #2765, opened 2024-12-09
Import action_view to support slim5

### LINKED ISSUE #2764: Does not work with slim
Only works if downgrading slim back to 4.  Unclear if the issue is with Middleman or Slim.  See https://stackoverflow.com/questions/15626845/how-to-install-and-use-slim-template-engine-with-middleman

jubishop:
this closes #2764

also see #2619 for more discussion.

### DIFF
--- a/middleman-core/lib/middleman-core/renderers/slim.rb
+++ b/middleman-core/lib/middleman-core/renderers/slim.rb
@@ -1,5 +1,6 @@
 # Load gem
 require 'slim'
+require 'action_view' # Required for Slim 5
 
 module SafeTemplate
   def render(*)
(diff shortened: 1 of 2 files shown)


jubishop:
bundle exec rake test seemed to not have any new failures beyond what was already failing on my machine.

I wasn't sure if I should also run `bundle install` and commit the new Gemfile.lock,  lmk.

markets:
Changes in `Gemfile.lock` shoud be committed too, but I'm not sure about this solution 🤔. If `actionview` is now a `Slim` dependency, shouldn't this come with the `Slim` gem by itself? I'm a bit confused.

markets:
Superseded by #2777

PR #2772, opened 2025-01-02
Relax tilt dependency

deivid-rodriguez:
Hello!

Not sure if the 5.x branch is actively maintained, but I found that 6a9b5c9cf2f5cc667ad539f5ac783ea0db95bdd9 made the dependency more strict.

However, I'm not sure if that was actually intended because the same commit is not present in the main branch.

So I though I'd propose this since it makes the dependency the same in both branches and allows using 5.x with Ruby 3.4 with less warnings (otherwise there are warnings fixed in the latest tilt version).


markets:
Hello @deivid-rodriguez 👋🏼 and thanks for this! 

The branch v5.x is probably abandoned, so it will be better to send patches to `main` (which is v4.x).

deivid-rodriguez:
Yeah, the thing is... this patch is not necessary in main, since that's what the dependency looks like in there.

I should probably try to start using main instead but sounded like it could be a non trivial upgrade, so I tried getting 5.x updated first.

deivid-rodriguez:
Oh, I'm confused, so I should be actually downgrading middleman? I thought main would be 6.x, not 4.x.

markets:
Actually, in `main` we still have this pinned version:

https://github.com/middleman/middleman/blob/c1cf7281b02eae165a6713079a3cba5660ec77aa/middleman-core/middleman-core.gemspec#L22

MM v5 is probably abandoned (#2400) and MM v6 is not planned at this point, our idea is to keep the v4.x active, but no big plans for now, just maintenance and upgrades, and things like that.

deivid-rodriguez:
Thanks for the info! Not sure where I got the idea that tilt was not pinned in main 😕. I'll reopen this PR targeting main.

deivid-rodriguez:
Done! And I also created #2773 to relax haml similarly.

deivid-rodriguez:
So it seems this upgrade nor the haml is not straightforward, according to the failed tests?

markets:
> So it seems neither this upgrade nor the haml one are straightforward, according to the failed tests?

Yep, it seems more tweaks are necessary.

deivid-rodriguez:
Just pushed some changes hoping they will fix things.

deivid-rodriguez:
I think this change also allows Slim 5 to be resolved with middleman, so I'm also removing the `slim < 5` pin from the Gemfile because I think it should work just fine now.

deivid-rodriguez:
Actually more work would be needed since tilt 2.5.0 is not supported and that's what's resolved if we allow Slim 5 in the development Gemfile, so I removed the last commit. I can work on that next.

markets (approved):
Thanks @deivid-rodriguez! Now all CI is green ✅ 

>  I can work on that next.

Please 👌🏼, any kind of help here is really welcome!

deivid-rodriguez:
This PR needs https://github.com/middleman/middleman/pull/2774 as a follow up, because tilt 2.5 is now allowed but doesn't really work.

PR #2773, opened 2025-01-02
Relax haml dependency to allow haml 5 and 6

deivid-rodriguez:
Haml 6 is supported by middleman's main branch, however, it became uninstallable since https://github.com/middleman/middleman/commit/6bf52366abf0ca89549e38b4873b701c66032f44.

Instead, the dependency from `middleman.gemspec` should've been preserved.


markets (approved):
Thanks again @deivid-rodriguez!

PR #2776, opened 2025-01-03
Support rack 3

deivid-rodriguez:
And set it as minimum required rack version, because it's not straightforward to support both rack 2 and rack 3.


deivid-rodriguez:
This breaks a few tests, so needs more work.

deivid-rodriguez:
Alright, this one should be ready now as well!

markets (approved):

PR #2777, opened 2025-01-03
Support slim 5

deivid-rodriguez:
Some tests are broken, but opening this as a WIP.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,6 @@ gem 'liquid', RUBY_VERSION <= '2.7' ? '< 5.6' : '> 0', require: false
 gem 'redcarpet', '>= 3.1', require: false
 gem 'sinatra', '~> 4.0', require: false
 gem 'slim', '~> 5.0', require: false
-gem 'tilt', '< 2.2', require: false # 2.2 or higher does not yet work with slim 5
 gem 'actionview', '~> 7.0', require: false
 
 # Dns server to test preview server
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,8 @@ gem 'pry-rescue'
 gem 'liquid', RUBY_VERSION <= '2.7' ? '< 5.6' : '> 0', require: false
 gem 'redcarpet', '>= 3.1', require: false
 gem 'sinatra', '~> 4.0', require: false
-gem 'slim', '~> 4.0', require: false
+gem 'slim', '~> 5.0', require: false
+gem 'tilt', '< 2.2', require: false # 2.2 or higher does not yet work with slim 5
 gem 'actionview', '~> 7.0', require: false
 
 # Dns server to test preview server
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ gem 'liquid', RUBY_VERSION <= '2.7' ? '< 5.6' : '> 0', require: false
 gem 'redcarpet', '>= 3.1', require: false
 gem 'sinatra', '~> 4.0', require: false
 gem 'slim', '~> 4.0', require: false
+gem 'actionview', '~> 7.0', require: false
 
 # Dns server to test preview server
 gem 'rubydns', '~> 1.0.1', require: false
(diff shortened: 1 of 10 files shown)


deivid-rodriguez:
Ok, I got tests passing with Slim 5 with the gotcha that tilt needs to be pinned to < 2.2. I can work on supporting tilt 2.2 next. For now, this PR fixes tests with Slim 5 (if not using tilt >= 2.2), and provides an alternative to #2765 that does not require adding `action_view` as a runtime dependency.

deivid-rodriguez:
This PR supersedes #2779 and #2765.

deivid-rodriguez:
For now I completely dropped support for tilt 2.1, because it's not straightforward to keep supporting new and old versions of tilt. Is it ok to drop tilt 2.1 support?

markets (approved):
Seems fine to me ✅ (and passes tests for all supported versions).

Thank you so much @deivid-rodriguez 🙌

markets:
**NOTE** We should check if all [renderers](https://github.com/middleman/middleman/tree/main/middleman-core/lib/middleman-core/renderers) are still supported in tilt 2.2+. I think they dropped support for Less, but not sure about the current status of Less, maybe we should remove it too?

deivid-rodriguez:
Good point, @markets!

Indeed `less` support was dropped in tilt 2.2: https://github.com/jeremyevans/tilt/commit/5f9381d9a87eac2df9c9a2cd195e3c6504e1a41e. Tests did not notice it so I guess there's no test coverage for less templates.

Less seems hard to use with Ruby, as explained in the tilt [commit message](https://github.com/jeremyevans/tilt/commit/e605a964e8fe40ea05f0eb5f4f490c6d3bc335b5) that deprecated it:

> Less on CRuby depends on therubyracer, which no longer works on modern Ruby due to changes in Psych.  I tried enabling it in CI, and it won't install (https://github.com/jeremyevans/tilt/runs/7812514562?check_suite_focus=true#step:4:155). Less hasn't had a release in over 8 years, and therubyracer hasn't had a release in over 5 years.

So I don't think it would make sense to keep supporting it in middleman. Tilt has been warning it already for a while, so it seems the natural direction to take.

Should I remove the renderer in this PR?

markets:
> Should I remove the renderer in this PR?

This seems the best decision in my opinion. It seems almost impossible to run Less in Ruby projects nowadays. We'll need to remove the renderer and this block too:

https://github.com/middleman/middleman/blob/cdda1d936194cb65a187ee87056d7ac44d305584/middleman-core/lib/middleman-core/core_extensions/rendering.rb#L45-L49

@tdreyno Are you also ok with this?

deivid-rodriguez:
Cool, I'll wait for @tdreyno's confirmation and then make the requested changes 👍.

markets:
@tdreyno sorry for the ping 🙏🏼 Are you ok to remove Less support? This PR fully fixes the Slim support by upgrading the Tilt engine, but as a side-effect removes Less support.

deivid-rodriguez:
Copilot left a rubocop-like suggestion, so I assume there was no need to address it since RuboCop was removed recently. But I addressed it just in case since it was a good one after all.

tdreyno:
@markets @deivid-rodriguez I'm okay with this change. Let's at least bump a minor release number.

I believe this can be replaced with a `external_pipeline` and the modern less CLI. Can we, instead of removing the template registration, have it throw an error message in the body of that template handler to recommend this approach with these docs: https://middlemanapp.com/advanced/external-pipeline/

https://github.com/middleman/middleman/pull/2777/files#diff-42862df91def5508ee0dec773d581819cdad5197529dc6a91868d0c469fea576L46-L49

deivid-rodriguez:
Ok, I'll try to implement that!

deivid-rodriguez:
@tdreyno I tried to implement your request 👍.

markets:
Nice @deivid-rodriguez 👌🏼

@tdreyno I agree with this statement:

> Let's at least bump a minor release number

After merging this PR, I think we can start preparing a v4.6 release. Current diff: https://github.com/middleman/middleman/compare/v4.5.1...main.

tdreyno (approved):

Issue #2788, opened 2025-01-16
`middleman` CLI fails on fresh install of Middleman


jvergeldedios:
I did a fresh install of `rbenv` and `ruby` to test this.

- OS: macOS 14.5
- Middleman version: 4.5.1
- Ruby version: 3.4.1
- Rubygems version: 3.6.2
- Bundler version: 2.6.3

Steps to reproduce:
- `gem install middleman`
- `middleman init projectname`

Results in the following stacktrace:

```
/Users/julian/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/activesupport-7.0.8.7/lib/active_support/logger_thread_safe_level.rb:12:in '<module:LoggerThreadSafeLevel>': uninitialized constant ActiveSupport::LoggerThreadSafeLevel::Logger (NameError)

    Logger::Severity.constants.each do |severity|
    ^^^^^^
	from /Users/julian/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/activesupport-7.0.8.7/lib/active_support/logger_thread_safe_level.rb:9:in '<module:ActiveSupport>'
	from /Users/julian/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/activesupport-7.0.8.7/lib/active_support/logger_thread_safe_level.rb:8:in '<top (required)>'
	from <internal:/Users/julian/.rbenv/versions/3.4.1/lib/ruby/3.4.0/rubygems/core_ext/kernel_require.rb>:136:in 'Kernel#require'
	from <internal:/Users/julian/.rbenv/versions/3.4.1/lib/ruby/3.4.0/rubygems/core_ext/kernel_require.rb>:136:in 'Kernel#require'
	from /Users/julian/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/activesupport-7.0.8.7/lib/active_support/logger_silence.rb:5:in '<top (required)>'
	from <internal:/Users/julian/.rbenv/versions/3.4.1/lib/ruby/3.4.0/rubygems/core_ext/kernel_require.rb>:136:in 'Kernel#require'
	from <internal:/Users/julian/.rbenv/versions/3.4.1/lib/ruby/3.4.0/rubygems/core_ext/kernel_require.rb>:136:in 'Kernel#require'
[...]

PR #2789, opened 2025-01-17
Add concurrent-ruby-1.3.4 as a dependency to address issues with activesupport

jvergeldedios:
As mentioned in https://github.com/middleman/middleman/issues/2788 a fresh install of Middleman breaks when trying to use the CLI to create a new project. This PR adds `concurrent-ruby-1.3.4` as a dependency to `middleman-core` to address the issue as described in https://github.com/alphagov/tech-docs-gem/commit/a7dc7dc409ec127a6a951f5c1d9c11833aa9a86a

Verified that this fixes the issue with the CLI on a locally built and installed gem.

### DIFF
--- a/middleman-core/middleman-core.gemspec
+++ b/middleman-core/middleman-core.gemspec
@@ -34,6 +34,7 @@ Gem::Specification.new do |s|
 
   # Helpers
   s.add_dependency('activesupport', ['>= 6.1'])
+  s.add_dependency("concurrent-ruby", "1.3.4")
   s.add_dependency('padrino-helpers', ['~> 0.15.0'])
   s.add_dependency("addressable", ["~> 2.4"])
   s.add_dependency('memoist', ['~> 0.14'])


markets:
I agree with @deivid-rodriguez, this doesn't seem a MM responsability.

jvergeldedios:
@deivid-rodriguez I'm confused about the `bundle update activesupport` part. This issue was fixed in activesupport https://github.com/rails/rails/pull/54264 but the particular version of activesupport that gets installed when you do `gem install middleman` is broken. The problem I ran into is specifically with the CLI which an end user doesn't have any way to update subdependencies via Bundler. So either MM needs to pin the subdependency of activesupport that is causing the issue (which is what this PR does) or require a newer version of activesupport than it currently selects on a fresh install.

deivid-rodriguez:
Ah I see @jvergeldedios. Then we should figure out what's pinning `activesupport` to the broken version, since this should be fixed in the latest `activesupport` version.

deivid-rodriguez:
Oh, I see, middleman used to have activesupport pinned until https://github.com/middleman/middleman/pull/2670. This just needs a middleman release including that PR in order to get fixed. And this PR should be reverted, since it will make middleman unusable for anyone not using exactly concurrent-ruby 1.3.4.

Fryguy:
@deivid-rodriguez Just an FYI that even though the issue is fixed in https://github.com/rails/rails/pull/54264, Rails core [doesn't want to release a new version with the fix](https://github.com/rails/rails/pull/54264#issuecomment-2596106532).

deivid-rodriguez:
But that's for 7-0-stable, right? Does Rails 8 also have the issue?

markets:
Yes @deivid-rodriguez! but Rails v8 (Active Support v8) works only for Ruby 3.2+ and we still support Ruby 2.7+. 

That change must have broken a lot of CIs!

I think it should be fixed upstream, it doesn't make a lot of sense 😅 to force a whole community of library maintainers to pin to specific versions and make changes to fix some incompatibilities with gems that we don't even use.

deivid-rodriguez:
I see. I proposed what I think it's a better workaround that middleman can provide at https://github.com/middleman/middleman/pull/2793.

PR #2793, opened 2025-01-23
Workaround activesupport issue

deivid-rodriguez:
The missing logger require is fixed in activesupport v8 or higher. And middleman can be used with Rails 8 as long as Ruby 3.2 or higher is used.

So add the missing logger dependency whenever activesupport is not modern enough, and old Ruby is used.

Users of more modern rubies may still run into the issue, but in that case the preferred fix is to upgrade activesupport.

### DIFF
--- a/middleman-core/lib/middleman-core/util.rb
+++ b/middleman-core/lib/middleman-core/util.rb
@@ -1,3 +1,9 @@
+if RUBY_VERSION < "3.2"
+  require "active_support/version"
+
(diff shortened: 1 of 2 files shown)


markets (approved):
Looks like a better approach for me 👍🏼

@jvergeldedios Could you please try this branch?

deivid-rodriguez:
There's easy way to try this branch when installing with `gem install` (it does not support installing from git sources), so I guess try it by directly editing the source code of the installed copy (it's what I did already anyways).

Issue #2666, opened 2023-10-28
 Missing Thor class for invoke middleman:generator


voytee:
## Expected behavior and actual behavior

 `Missing Thor class for invoke middleman:generator` raised when doing `middleman init`

## Additional information

- Ruby version: 3.2.2
- Middleman version: 4.5.0
- OS version: OSX

tdreyno:
Can you try on 4.5.1?

voytee:
> Can you try on 4.5.1?

```gem install middleman -v 4.5.1
ERROR:  Could not find a valid gem 'middleman' (= 4.5.1) in any repository```

Seems like 4.5.1 is not available through ruby gems and I do not see 4.5.1 TAG on the repo

**Comment by voytee** [comment]
Also checked with `5.0.0.rc.1` but with this one you'll get other error:

```uninitialized constant Haml::Options```

**Comment by tdreyno** [comment]
4.5.1 is definitely on rubygems: https://rubygems.org/gems/middleman/versions/4.5.1

**Comment by arashm** [comment]
When doing `gem install middleman`, rubygems installs the `4.5.0` version. Forcing the version also fails. 

```
─$ gem install middleman -v 4.5.1
ERROR:  Could not find a valid gem 'middleman' (= 4.5.1) in any repository
ERROR:  Possible alternatives: beans-middleman, branchable_cdn_assets-middleman, capistrano-middleman, capistrano-middleman-presentation, capistrano-middleman2, contentful-middleman-pages, contentful_middleman, ember-middleman, fdt_middleman_tools, font-awesome-middleman
```

It could be rubygems issue though.

nicolas-brousse:
I've got the same error. By using an older version of `thor` (`1.2.2`) instead `1.3.0` the `middleman init` command works.

heyrocker:
Any updates on this? I also am having this problem and can't get 4.5.1 to install.

tdreyno:
Seems like 474 people have successfully installed 4.5.1
<img width="994" alt="Screenshot 2023-11-15 at 10 05 20" src="https://github.com/middleman/middleman/assets/233/745fab7b-7dc2-4615-8a17-88de6272a03f">

I may just have to republish and hope that ruby gems stops being confused

markets:
@tdreyno I successfully installed the v4.5.1 release for my projects under Ruby 3.2. 

I miss the git tag in the repository, they are useful to compare changes between versions: https://github.com/middleman/middleman/tags. 

[...]

Issue #2665, opened 2023-10-19
Issue creating (init) new site with thor 1.3.0


ttych:
## Expected behavior and actual behavior

From a fresh install of middleman,
I am not able to create a new site with **init** command.
There is an error between Thor and middleman.

I put all the details in the following parts.
I did not get what changed between 1.2.2 and 1.3.0 in Thor to explain this error now.

**Many thanks !**

## Steps to reproduce the problem (from a clean middleman installation)

From a fresh ruby install (with rbenv), install middleman

``` shell
$ gem install middleman
Fetching execjs-2.9.1.gem                                                      
Fetching servolux-0.13.0.gem                                                   
Fetching uglifier-3.2.0.gem                                                    
Fetching tilt-2.0.11.gem                                                       
Fetching toml-0.3.0.gem                                                        
Fetching webrick-1.8.1.gem                                                     
Fetching ffi-1.16.3.gem                                                                                                                                        
Fetching parslet-2.0.0.gem                                                     
Fetching sassc-2.4.0.gem                                                       
Fetching rack-2.2.8.gem                                                        
Fetching parallel-1.23.0.gem                                                   
Fetching padrino-support-0.15.3.gem                                            
Fetching concurrent-ruby-1.2.2.gem                                             
Fetching i18n-1.6.0.gem                                                        
Fetching padrino-helpers-0.15.3.gem                                            
Fetching memoist-0.16.2.gem                                                    
Fetching rb-inotify-0.10.1.gem                                                 
Fetching rb-fsevent-0.11.2.gem                                                 
Fetching listen-3.8.0.gem                                                      
Fetching hashie-3.6.0.gem                                                      
Fetching hamster-3.0.0.gem                                                     
Fetching fastimage-2.2.7.gem                                                   
Fetching fast_blank-1.0.1.gem                                                  
Fetching erubis-2.7.0.gem                                                      
Fetching dotenv-2.8.1.gem                                                      
Fetching contracts-0.17.gem                                                    
Fetching backports-3.24.1.gem                                                  
Fetching public_suffix-5.0.3.gem                                               
Fetching addressable-2.8.5.gem                                                 
Fetching tzinfo-2.0.6.gem                                                      
Fetching activesupport-7.0.8.gem                                               
Fetching middleman-core-4.5.0.gem                                              
Fetching thor-1.3.0.gem                                                        
Fetching middleman-cli-4.5.0.gem                                               
Fetching kramdown-2.4.0.gem                                                    
[...]

PR #2811, opened 2025-03-11
Allow upgrade to thor 1.3.x by correcting `init` behaviour

chadlwilson:
In https://github.com/middleman/middleman/commit/ad0e0ee9ba5e017e4f3f1cc861f4fa2a4c04f198 thor was locked at `< 1.3.0` due to #2665 and #2666 .

At root, the issue is because of https://github.com/rails/thor/releases/tag/v1.3.0 and the changes in https://github.com/rails/thor/pull/754 (and related commits such as https://github.com/rails/thor/commit/a601ca9342729350529ee8235abdf9447e77474e) regarding the way Thor 1.3 locates namespaces/classes/commands.

What happens currently is that Thor 1.3 gets completely confused and determines `klass = nil`, `command = middleman`, hence the error `'invoke': Missing Thor class for invoke middleman:generator (RuntimeError)`.

It seems to be that the namespace/class location logic no longer works the same way if the `invoke` instruction does not have the task/command included. Perhaps the previous logic assumed if there was only a single function/task that it would select it.

There are two possible solutions/workarounds I've found
1. Update all templates to declare `namespace :middleman` explicitly
2. The change here, to explicitly include the `command`.

Both options seem backward compatible, at least with thor 1.2.x. I chose the second, as it seemed more straightforward than updating all of the template repos, and seems more explicit.and the task to run.

I'm not enough of a thor export to know if this change was intentional on thor side, or if middleman was unintentionally relying on undocumented thor behavior of invoke where one doesn't explicitly note the command.

Templates
- https://github.com/middleman/middleman-templates-default/blob/master/Thorfile
- https://github.com/middleman/middleman-templates-blog/blob/master/Thorfile
- https://github.com/middleman/middleman-templates-default-v5/blob/master/Thorfile

If the maintainers think that this is a bug in the way thor now works I can instead look at fixing upstream instead.

**Testing**
- Have also run the GHA tests across Ruby versions with the existing latest thor 1.2.2 version

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: middleman-cli
   specs:
     middleman-cli (4.6.0)
-      thor (>= 0.17.0, < 1.3.0)
+      thor (>= 0.17.0, < 2)
 
 PATH
   remote: middleman-core
@@ -266,7 +266,7 @@ GEM
     sys-uname (1.3.1)
       ffi (~> 1.1)
     temple (0.10.3)
-    thor (1.2.2)
+    thor (1.3.2)
     tilt (2.6.0)
     timers (4.0.4)
       hitimes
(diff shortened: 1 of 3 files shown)


tdreyno:
@chadlwilson this is fantastic. thanks for looking into this. Will merge if CI passes

chadlwilson:
Thanks for all your work @tdreyno !

PR #2839, opened 2025-07-17
Use lowercase headers in preview app

### LINKED ISSUE #2838: Middleman's settings are not accessible in a new application
Starting a new Middleman project and running middleman serve works, but accessing the settings page causes an internal server error. The issue seems related to Rack::Lint rejecting uppercase header names

**Steps to reproduce:**

```bash
ruby --version
# ruby 3.4.4 (2025-05-14 revision a38531fd3f) +PRISM [x86_64-linux]

middleman init .

middleman serve
```

```bash
curl http://localhost:4567/__middleman/config/
```

**Result:**

```html
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN">
<HTML>
  <HEAD><TITLE>Internal Server Error</TITLE></HEAD>
  <BODY>
    <H1>Internal Server Error</H1>
    uppercase character in header name: Content-Type
    <HR>
    <ADDRESS>
     WEBrick/1.9.1 (Ruby/3.4.4/2025-05-14) OpenSSL/3.5.1 at
     localhost:4567
    </ADDRESS>
  </BODY>
</HTML>
```

Is this related to the change in Middleman 4.6 that uses rackup with rack > 3?

deivid-rodriguez:
Since Rack 3 is more strict about it.

Fixes #2838.

### DIFF
--- a/middleman-core/lib/middleman-core/meta_pages.rb
+++ b/middleman-core/lib/middleman-core/meta_pages.rb
@@ -101,7 +101,7 @@ module Middleman
 
       # Respond to an HTML request
       def response(content)
-        [200, { 'Content-Type' => 'text/html' }, Array(content)]
+        [200, { 'content-type' => 'text/html' }, Array(content)]
       end
 
       def extension_options(extension)


deivid-rodriguez:
For context, we were already prepared for this since https://github.com/middleman/middleman/pull/2706, but the preview app wasn't updated.

PR #2857, opened 2025-09-17
Fix implicit 1x scale in srcset crashing middleman

imcrazytwkr:
I've stumbled upon a problem where if you write a syntaxically correct srcset for fixed size with implicit `1x` descriptor that can be omitted ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Responsive_images#resolution_switching_same_size_different_resolutions)), middleman crashes when opening the page because ruby cannot concat string with `nil` and throws an error. This fix allows for implicit descriptors in `srcset` while using `image_tag`.


markets:
Hello @imcrazytwkr 👋🏼

Yes, that was a flaky test for ages... let me re-run that one.

markets:
Thanks @imcrazytwkr 🙌🏼

PR #2880, opened 2025-11-18
fix syntax error in kramdown.rb

beewill:
```
undefined local variable or method `context' for an instance of Middleman::Renderers::KramdownTemplate (NameError)

        MiddlemanKramdownHTML.scope = @context || context
                                                  ^^^^^^^
```


markets:
Thanks @beewill 🤝

markets (approved):

PR #2896, opened 2025-12-27
Drop support for EOL Ruby versions

mmenanno:
I think preventing current dependancies versions solely because based on supporting EOL Ruby versions might be a code smell (https://github.com/middleman/middleman/pull/2877). Dropping support for the EOL versions of Ruby, and in turn relaxing the HAML versions restrictions again.

I updated the matrix for CI ruby versions but someone would need to change the repo settings to change which ones are set to "required"

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


chadlwilson:
Thanks for this @mmenanno !

